PR #21, opened 2020-07-08
Umbrel v0.1.3-beta

mayankchhabra:
After merge:
- [ ] Tag and release Umbrel v0.1.3-beta

### DIFF
--- a/install-box.sh
+++ b/install-box.sh
@@ -13,7 +13,7 @@
 
 echo "Cloning to current working directory from github..."
 git init
-git remote add origin https://github.com/mayankchhabra/umbrel.git
+git remote add origin https://github.com/getumbrel/umbrel.git
 git fetch --all --tags
 git checkout tags/v0.1.3-beta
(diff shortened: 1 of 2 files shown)


install-box.sh
```diff
@@ -12,7 +12,11 @@
 # Pre-requisites: git
 
 echo "Cloning to current working directory from github..."
-git clone -b v0.1.2 https://github.com/getumbrel/umbrel.git .
+git init
+git remote add origin https://github.com/mayankchhabra/umbrel.git
```
mayankchhabra, line 16: Change to official repo after testing

nolim1t:
You might want to cherry pick some commits from https://github.com/getumbrel/umbrel/pull/22 and test it if you want

mayankchhabra:
This is only a patch release, I think #22 deserves it's own minor release, plus some updates to `umbrel-dashboard` will also need to be made where it's explicitly stated in the 'node info' section that lnd automatically switches between neutrino and bitcoind.

PR #22, opened 2020-07-08
Neutrino switch container

### LINKED ISSUE #2: Neutrino 🔁 Bitcoin Node Automatic Switching
### Issue

- The user cannot start using his lightning or bitcoin wallet until `bitcoind` has fully synchronized. This adds a lot of friction due to the long delay, creating a very frustrating experience.

- If `bitcoind` lags behind for some reason, such as when the user had his node turned off for a few days/weeks, the same problem reoccurs.

### Solution

Neutrino filters can solve this problem. We need an automatic, recurring way to switch lnd's bitcoin mode to `neutrino` from `bitcoind` as soon as `bitcoind` lags behind by a certain number of blocks from its best header.

### Ideas

Perhaps it's a good idea to create a new service that runs 24x7 via a container called `lnd-switch`  that uses the following environment variables:

```
BITCOIN_HOST: "0.0.0.0"
BITCOIN_RPC_PORT: "8332"
BITCOIN_RPC_USER: "umbrel"
BITCOIN_RPC_PASSWORD: "1234567890"
LND_CONFIG_FILE: "/home/umbrel/lnd/lnd.conf"
SWITCH_TO_NEUTRINO: "144" (switch to neutrino when bitcoind is *at least* 144 blocks behind)
SWITCH_TO_BITCOIND: "3" (switch to bitcoind when bitcoind is *at max* 3 blocks behind)
```

Considering the above example variables, `lnd-switch` will update the bitcoin node in `LND_CONFIG_FILE`  from `bitcoin.node=bitcoind` to `bitcoin.node=neutrino` as soon as `bitcoind`'s `headers - blocks >= 144` and update it from `bitcoin.node=neutrino` to `bitcoin.node=bitcoind` as soon as `bitcoind`'s `headers - blocks <= 3`. It also needs a way to restart the `lnd` container after every config file change.

I'm not sure if this is the right approach or not. If it is, then `lnd-switch` can be a tiny FOSS project on its own that (I think) many more people will find useful.

cc @nolim1t @meeDamian

nolim1t:
Closes #2 when resolved

Currently this will notify that it has switched configurations (in form of a file) but it will not restart the node for the sake of not doing stuff without telling the user. From a users point of view they need not care if it is neutrino or not.  I've designed it in a way to be non-intrusive.

Eventually we can have the dashboard notify of important updates and details of said update

Also if the node is pruned it will continue to use neutrino as there may be issues with pruned nodes and LND when the channel creation date is no longer in the node history. So this can ONLY be tested properly on full nodes. It will continue to work seamlessly on any nodes but won't do anything.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -129,6 +129,7 @@ services:
                     net:
                         ipv4_address: 10.11.2.2
         neutrino-switcher:
+                container_name: neutrino-switcher
                 image: lncm/neutrino-switcher:1.0.3
                 logging: *default-logging
                 depends_on: [ bitcoin, lnd ]
@@ -137,8 +138,11 @@ services:
                     - "${PWD}/lnd:/lnd"
                     - "${PWD}/secrets:/secrets"
                     - "${PWD}/statuses:/statuses"
+                    - "/var/run/docker.sock:/var/run/docker.sock"
                 environment:
                     JSONRPCURL: http://10.11.1.1:RPCPORT
+                    LND_CONTAINER_NAME: lnd
+                    SLEEPTIME: 86400
                 networks:
                     net:
                         ipv4_address: 10.11.5.2
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -130,7 +130,7 @@ services:
                         ipv4_address: 10.11.2.2
         neutrino-switcher:
                 container_name: neutrino-switcher
-                image: lncm/neutrino-switcher:1.0.3
+                image: lncm/neutrino-switcher:1.0.2
                 logging: *default-logging
                 depends_on: [ bitcoin, lnd ]
                 restart: always
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -125,6 +125,20 @@ services:
                 networks:
                     net:
                         ipv4_address: 10.11.2.2
+        neutrino-switcher:
+                image: lncm/neutrino-switcher:1.0.0
+                logging: *default-logging
+                depends_on: [ bitcoin, lnd ]
+                restart: always
+                volumes:
+                    - "${PWD}/lnd:/lnd"
+                    - "${PWD}/secrets:/secrets"
+                    - "${PWD}/statuses:/statuses"
+                environment:
+                    JSONRPCURL: http://10.11.1.1:RPCPORT
+                networks:
+                    net:
+                        ipv4_address: 10.11.5.2
 networks:
     net:
         ipam:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -133,7 +133,7 @@ services:
                 image: lncm/neutrino-switcher:1.0.2
                 logging: *default-logging
                 depends_on: [ bitcoin, lnd ]
-                restart: always
+                restart: unless-stopped
                 volumes:
                     - "${PWD}/lnd:/lnd"
                     - "${PWD}/secrets:/secrets"
@@ -142,7 +142,7 @@ services:
                 environment:
                     JSONRPCURL: http://10.11.1.1:RPCPORT
                     LND_CONTAINER_NAME: lnd
-                    SLEEPTIME: 86400
+                    SLEEPTIME: 3600
                 networks:
                     net:
                         ipv4_address: 10.11.5.2
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -129,7 +129,7 @@ services:
                     net:
                         ipv4_address: 10.11.2.2
         neutrino-switcher:
-                image: lncm/neutrino-switcher:1.0.0
+                image: lncm/neutrino-switcher:1.0.3
                 logging: *default-logging
                 depends_on: [ bitcoin, lnd ]
                 restart: always
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ If you are looking to run Umbrel on your hardware, you do not need to use this r
 - [`LND`](https://github.com/getumbrel/docker-lnd)
 - [`Tor`](https://github.com/getumbrel/docker-tor)
 - [`Nginx`](https://github.com/nginx/nginx)
+- [`lnd-neutrino-switch`](https://github.com/lncm/docker-lnd-neutrino-switch) - Utility container for seamlessly transitioning from neutrino mode to bitcoind mode
 
 
 **Architecture**
--- a/lnd/lnd.conf
+++ b/lnd/lnd.conf
@@ -38,7 +38,6 @@ bitcoind.estimatemode=ECONOMICAL
 ; If the Bitcoin chain should be active. Atm, only a single chain can be
 ; active.
 bitcoin.active=1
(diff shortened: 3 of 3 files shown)


mayankchhabra:
Great job, @nolim1t! Will try to test this out today. However, why can't we restart the lnd node? We can explicitly state in the "node info" section that lnd will automatically switch back and forth between bitcoind and neutrino depending upon sync. 

If lnd is not restarted, then there's no point in changing the `lnd.conf` file because it's never loaded (until of course the next restart).

Eventually down the lane, and to make it more clear for the user, we should add a new "backend" option on the lightning page (in "node info' section) with a 3-selection toggle button that looks something like this:

image

 The three options will be:

- Bitcoin Core
- Neutrino
- Auto

By default, we can ship with "auto" pre-selected, which means lightning will automatically switch back and forth between neutrino and bitcoind depending upon bitcoin's sync. 

If the node is pruned, the "Bitcoin Core" option will be disabled.

If the user selects "Neutrino", then we should not switch back and forth and use neutrino at all times.

For users who don't care/understand, "Auto" will do its job.

nolim1t:
If you can figure out a way to stop and start the LND container from another container.

mayankchhabra:
We can do something like this:

- Mount the docker socket from host to `lncm/neutrino-switcher:1.0.0` using `/var/run/docker.sock:/var/run/docker.sock`
- Pass env variable called LND_CONTAINER_NAME with default value "lnd"
- After updating the `lnd.conf`, run `docker restart $LND_CONTAINER_NAME`

nolim1t:
interesting..  ```docker``` commands runs within containers lol

mayankchhabra:
Also need to bind mount the docker binary `/usr/bin/docker:/usr/bin/docker`

We already restart the `lnd` container from `umbrel-manager` container after the password is changed :)
https://github.com/getumbrel/umbrel/blob/826573ed27a70379c4a61dc23268ed9ca2dca1c2/docker-compose.yml#L88

nolim1t:
Actually ... I can just install the docker package into the container and it seems to so far work hahaha

mayankchhabra:
Yes, but you don't need to because there's not much benefit to it. Rather:

- It will increase the image size of `lncm/neutrino-switcher`.
- There can be a mismatch of docker version inside `lncm/neutrino-switcher` and on the host, where the one inside the image can be incompatible with the docker socket of the host. 

Since we're using docker socket of the host, it's preferable to use host's docker binary too

nolim1t:
Having the docker binary within the container works fine, as its all cross compiled using buildx. We need to test this on a variety of environments thoroughly.

I'll be releasing ```1.0.1``` of the container with all the latest changes after i've tested it.

nolim1t:
The container will restart LND automatically. The docker binary seems to be able to control stuff from the container ok enough

nolim1t:
Also updated the README file in here

mayankchhabra:
Ok I tried to run this on a clean install, and one of the things that we need to do it is to have `neutrino-switcher` run only when both `bitcoind` and `lnd` are running as well as available. For example, I got this error on `docker-compose up`:

```
neutrino-switcher    | Checking if synced...
neutrino-switcher    | If set to neutrino then lets check bitcoind
neutrino-switcher    | Error: from bitcoind
neutrino-switcher    | {
neutrino-switcher    |   "code": -28,
neutrino-switcher    |   "message": "Loading wallet..."
neutrino-switcher    | }
```

We can't solve it using `depends_on` in `docker-compose.yml` because it only checks for the container to be running, not for the service inside it to be available.

Now the last time we had this problem when `umbrel-middleware` needed `umbrel-manger` to be available before starting up, we did a custom command that used this [script](https://github.com/getumbrel/umbrel-middleware/blob/master/wait-for-node-manager.sh) when starting the middleware:

https://github.com/getumbrel/umbrel/blob/e25737712ad3c4560c9de9f2af94409e57ace673/docker-compose.yml#L112

So something like this needs to be implemented in the `neutrino-switcher` container too.

nolim1t:
No that is fine.. it will just check on next pass.

nolim1t:
You can speed up the checking. It is configurable.

I just kept things conservative to minimize restarts, as it doesn't actually affect any operational stuff. The key thing to look out for is if LND cares enough if things are just switched out. I don't think its an issue though.

docker-compose.yml
```diff
@@ -128,6 +128,24 @@ services:
...
+                restart: always
+                volumes:
+                    - "${PWD}/lnd:/lnd"
+                    - "${PWD}/secrets:/secrets"
+                    - "${PWD}/statuses:/statuses"
+                    - "/var/run/docker.sock:/var/run/docker.sock"
+                environment:
+                    JSONRPCURL: http://10.11.1.1:RPCPORT
+                    LND_CONTAINER_NAME: lnd
+                    SLEEPTIME: 86400
```
mayankchhabra, line 145: Can we make it check every hour? 1 day is a loooong time! The computational overhead to check every hour is negligible.

docker-compose.yml
```diff
@@ -128,6 +128,24 @@ services:
                 networks:
                     net:
                         ipv4_address: 10.11.2.2
+        neutrino-switcher:
+                container_name: neutrino-switcher
+                image: lncm/neutrino-switcher:1.0.2
+                logging: *default-logging
+                depends_on: [ bitcoin, lnd ]
+                restart: always
```
mayankchhabra, line 136: Also please change the restart policy to `unless-stopped`. All containers should stop when `umbrel-manager` (or anyone else) tries to stop them.

nolim1t:
You can do it whatever you wantOn Fri, Jul 10, 2020 at 17:56, Mayank Chhabra <notifications@github.com> wrote:  
@mayankchhabra commented on this pull request.



In docker-compose.yml:
> @@ -128,6 +128,24 @@ services:
                 networks:
                     net:
                         ipv4_address: 10.11.2.2
+        neutrino-switcher:
+                container_name: neutrino-switcher
+                image: lncm/neutrino-switcher:1.0.2
+                logging: *default-logging
+                depends_on: [ bitcoin, lnd ]
+                restart: always
+                volumes:
+                    - "${PWD}/lnd:/lnd"
+                    - "${PWD}/secrets:/secrets"
+                    - "${PWD}/statuses:/statuses"
+                    - "/var/run/docker.sock:/var/run/docker.sock"
+                environment:
+                    JSONRPCURL: http://10.11.1.1:RPCPORT
+                    LND_CONTAINER_NAME: lnd
+                    SLEEPTIME: 86400

Can we make it check every hour? 1 day is a loooong time! The computational overhead to check every hour is negligible.

docker-compose.yml
nolim1t: Done

docker-compose.yml
nolim1t: 1 hour now

mayankchhabra:
Awesome work @nolim1t! Merging now :D

PR #23, opened 2020-07-09
⚠️ Safety First ⚠️

nolim1t:
Added dependency checking to bash scripts

### DIFF
--- a/configure-box.sh
+++ b/configure-box.sh
@@ -11,6 +11,17 @@
 # Install the docker-compose box to the current working directory
 # Pre-requisites: wget
 
+check_dependencies () {
+  for cmd in "$@"; do
+    if ! command -v $cmd >/dev/null 2>&1; then
+      echo "This script requires \"${cmd}\" to be installed"
+      exit 1
+    fi
+  done
+}
+
+check_dependencies wget docker
+
 echo "Start box configuration"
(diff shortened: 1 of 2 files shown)


configure-box.sh
```diff
@@ -11,6 +11,17 @@
...
+check_dependencies () {
+  for cmd in "$@"; do
+    if ! command -v $cmd >/dev/null 2>&1; then
+      echo "This script requires \"${cmd}\" to be installed"
+      exit 1
+    fi
+  done
+}
+
+check_dependencies wget docker
```
mayankchhabra, line 23: Can you also check for `docker-compose` here?

nolim1t: Actually docker-compose isn't required to run the install. Everything can be set up without it.

The main purpose of this is to make it easy to install and configure.

nolim1t:
FYI.. the checks are only whats needed to complete the task at hand.  That is installing  and configuring.

mayankchhabra:
Yep, `docker-compose pull` is run inside `configure-box.sh`
https://github.com/getumbrel/umbrel/blob/b47d1cb16d9e62bd8c48977db13870a65b6f38af/configure-box.sh#L80

nolim1t:
Added while resolving the conflict

PR #24, opened 2020-07-09
Add a check for Operating systems

nolim1t:
Add a check for Operating systems as there are lot of differences between Linux and OS X, and we have only tested mainly on linux / debian based systems


mayankchhabra:
Tested. Great job @nolim1t! :)

PR #25, opened 2020-07-09
Replace git logic with a single git command

lukechilds:
This is functionally equivalent.

### DIFF
--- a/install-box.sh
+++ b/install-box.sh
@@ -12,11 +12,7 @@
 # Pre-requisites: git
 
 echo "Cloning to current working directory from github..."
-git init
-git remote add origin https://github.com/getumbrel/umbrel.git
-git fetch --all --tags
-git checkout tags/v0.1.3-beta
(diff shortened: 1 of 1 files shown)


lukechilds:
Actually it's not the same, a fresh clone will fail if the directory contains files.

Is think `git reset --hard tags/v0.1.3-beta` is still redundant though.

mayankchhabra:
Lol literally had to make an emergency release yesterday because I did the same mistake: https://github.com/getumbrel/umbrel/pull/21/files

lukechilds:
Do you know the reasoning behind `git reset --hard`?

mayankchhabra:
Not really, AFAIU you're right — it seems redundant since no local changes are made between `git checkout` and `git reset`. I just reverted it back to what it was. 

cc @nolim1t

nolim1t:
I'm not just "flattening the curve" .. I'm destroying it completely

no-flattnening

PR #28, opened 2020-07-13
Add signal watcher system with power cycle triggers

lukechilds:
Related:
- https://github.com/getumbrel/umbrel-manager/pull/21
- https://github.com/getumbrel/umbrel-dashboard/pull/130

**Edit:** The below description is outdated see https://github.com/getumbrel/umbrel/pull/28#issuecomment-658138242 for the latest update and instructions on how to test.

Adds a script called `signal-watcher` to Umbrel dir that when executed will take a param for a dir to watch.

Inside that dir it will watch for files inside a `/signals` subdir, when one is modified it will check if there is an executable file of the same name in `/triggers`, if it exists it will execute it.

I think it's good to implement triggers as executables instead of directly inside the watcher script a few reasons:
- Easy to add new features, just add a new script file, no need to scroll though a load of if/else blocks
- Keeps logic nice and separated
- Doesn't limit us to bash, a trigger can be a bash script, a compiled binary, a python script, etc
- Doesn't kill the watcher script if something in the trigger fails
- Allows us to execute triggers async so a long running task doesn't slow down other triggers

`signal-watcher` should be run as root so it can perform privileged tasks such as shutdown.

To add a new signal all you need to do is create a new executable file in the `/triggers` subdir.

Demo:

```
$ ./signal-watcher watcher
```

Then in another shell:

```
$ touch signals/foo
```

Means first shell will log:

```
Got signal: foo
No trigger found
```

Then:

```
$ touch signals/shutdown
```

Means first shell will log:

```
Got signal: shutdown
Executing trigger
System is shutting down...
```


mayankchhabra:
Both concept and implementation ACK! Love it, very clean! 👏 

This will also work with the OTA update implementation right away. Quick question though, did you deliberately skip adding `docker-compose down` to both `reboot` and `shutdown` triggers?

lukechilds:
> Quick question though, did you deliberately skip adding docker-compose down to both reboot and shutdown triggers?

I could be mistaken but I don't think that's necessary.

The Docker daemon should receive a shutdown signal from the system and then attempt to safely shutdown any running containers before exiting.

mayankchhabra:
Ah yes, that could work too. But I don't think it's going to respect the `docker-compose.yml` config, such as we might lose on the extended grace periods:
https://github.com/getumbrel/umbrel/blob/ab200d389adabceb345197e9a2356a449e4b15cd/docker-compose.yml#L50

Which will only take effect if we use `docker-compose` and not `docker`. Or am I missing something?

lukechilds:
I think it will still respect them since I think the grace period is actually set on the container when it's created, so Docker know however this container is shutdown, it should wait the grace period until it force kills it.

I'm not 100% sure about that but that makes sense to me and that's just how I assumed it worked. I will test it to make sure.

lukechilds:
To test you'll need to checkout my `power-cycle` branches of `umbrel`, `umbrel-manager`, and `umbrel-dashboard`. (see related PRs)

You'll also need to run `karen` as root for the system to actually respond to the power cycle triggers.

If you test without `karen` running at first, you can just `ls events/signals/` to watch the files be created. When `karen` isn't running the the UI will still all follow through as if the system was power cycling, but the system won't actually do anything.

mayankchhabra:
Tested on physical RPi and everything worked flawlessly. Karen is my favorite, haha! 
Great work, @lukechilds. 🙌 

Merging this now :)

PR #29, opened 2020-07-13
Partitioner fix

nolim1t:
Fix for partitioner where it fails if there is no partitions


mayankchhabra:
Awesome! Going to test this out today

mayankchhabra:
Tested and it worked! Great work, @nolim1t!

PR #33, opened 2020-07-15
Umbrel v0.1.4-beta

mayankchhabra:
Releasing this on a tight schedule so we can deploy partition fixes and shutdown/reboot functionality via a new Umbrel OS release immediately.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -69,7 +69,7 @@ services:
                         ipv4_address: 10.11.1.2
         dashboard:
                 container_name: dashboard
-                image: getumbrel/dashboard:v0.2.2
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -69,7 +69,7 @@ services:
                         ipv4_address: 10.11.1.2
         dashboard:
                 container_name: dashboard
-                image: getumbrel/dashboard:v0.2.2
+                image: getumbrel/dashboard:master
```
mayankchhabra, line 72: Change this to release tag after releasee

mayankchhabra:
Tested.

PR #35, opened 2020-07-17
Expose bitcoind rpc to TOR

nolim1t:
Expose bitcoind rpc tor TOR for https://github.com/getumbrel/umbrel-dashboard/issues/125

### DIFF
--- a/tor/torrc
+++ b/tor/torrc
@@ -6,5 +6,11 @@ ControlPort 10.11.5.1:29051
 HiddenServiceDir /var/lib/tor/web
 HiddenServicePort 80 10.11.0.2:80
 
+# Bitcoind RPC Hidden Service (Mainnet / Testnet / Regtest)
+HiddenServiceDir /var/lib/tor/bitcoind-rpc
+HiddenServicePort 8332 10.11.1.1:8332
+HiddenServicePort 18332 10.11.1.1:18332
+HiddenServicePort 18443 10.11.1.1:18443
+
 # Tor Passwords
 HashedControlPassword 16:50A873DF18C00F4A6048BF1CEF7E7AA66478F0B5134DA4369D80657F26


mayankchhabra:
Concept ACK. But we do not have to manually create hidden services for Bitcoin Core as it automatically creates them without any configuration required from our end (check `docker-compose logs -f bitcoin` and scroll to the top, you'd find `bitcoind` has printed it out somewhere).

To fix getumbrel/umbrel-dashboard#125, we just need to get that onion URL from Bitcoin Core using `umbrel-middleware`.

nolim1t:
Not rpc we can’t..The p2p port is only exposed and you don’t need to do it from logs.On Fri, Jul 17, 2020 at 23:49, Mayank Chhabra <notifications@github.com> wrote:  
Concept ACK. But we do not have to manually create hidden services for Bitcoin Core as it automatically creates them without any configuration required from our end (check docker-compose logs -f bitcoin and scroll to the top, you'd find bitcoind has printed it out somewhere).
To fix getumbrel/umbrel-dashboard#125, we just need to get that onion URL from Bitcoin Core using umbrel-middleware.

mayankchhabra:
Notice the `localaddresses` key. We just need to get that through `umbrel-middleware`.

```sh
$ docker exec -it bitcoin bitcoin-cli getnetworkinfo

{
  "version": 200000,
  "subversion": "/Satoshi:0.20.0/",
  "protocolversion": 70015,
  "localservices": "0000000000000408",
  "localservicesnames": [
    "WITNESS",
    "NETWORK_LIMITED"
  ],
  "localrelay": false,
  "timeoffset": 0,
  "networkactive": true,
  "connections": 0,
  "networks": [
    {
      "name": "ipv4",
      "limited": false,
      "reachable": true,
      "proxy": "",
      "proxy_randomize_credentials": false
    },
    {
      "name": "ipv6",
      "limited": false,
      "reachable": true,
      "proxy": "",
      "proxy_randomize_credentials": false
    },
    {
      "name": "onion",
      "limited": false,
      "reachable": true,
      "proxy": "10.11.5.1:29050",
      "proxy_randomize_credentials": true
    }
  ],
  "relayfee": 0.00001000,
  "incrementalfee": 0.00001000,
  "localaddresses": [
    {
      "address": "7pjrzotsklc6bb6g.onion",
      "port": 18444,
      "score": 4
    }
  ],
  "warnings": ""
}
```

nolim1t:
No that is the p2p portThe rpc port isn’t exposed to tor   
Notice the localaddresses key. We just need to get that through umbrel-middleware.
$ docker exec -it bitcoin bitcoin-cli getnetworkinfo

{
  "version": 200000,
  "subversion": "/Satoshi:0.20.0/",
  "protocolversion": 70015,
  "localservices": "0000000000000408",
  "localservicesnames": [
    "WITNESS",
    "NETWORK_LIMITED"
  ],
  "localrelay": false,
  "timeoffset": 0,
  "networkactive": true,
  "connections": 0,
  "networks": [
    {
      "name": "ipv4",
      "limited": false,
      "reachable": true,
      "proxy": "",
      "proxy_randomize_credentials": false
    },
    {
      "name": "ipv6",
      "limited": false,
      "reachable": true,
      "proxy": "",
      "proxy_randomize_credentials": false
    },
    {
      "name": "onion",
      "limited": false,
      "reachable": true,
      "proxy": "10.11.5.1:29050",
      "proxy_randomize_credentials": true
    }
  ],
  "relayfee": 0.00001000,
  "incrementalfee": 0.00001000,
  "localaddresses": [
    {
      "address": "7pjrzotsklc6bb6g.onion",
      "port": 18444,
      "score": 4
    }
  ],
  "warnings": ""
}

lukechilds:
@mayankchhabra Bitcoin's P2P communication and RPC communication are two completely different things with different ports. When Tor automatically configures the hidden service it only exposes the P2P port, not the RPC port.

@nolim1t's PR is creating a new hidden service specifically for RPC.

Are we sure we wan't to do this though? RPC is supposed to be private, not public. I know there are some wallets that support connecting to remote RPC over Tor but I'm not sure RPC is really designed or tested to be publicly facing.

Assuming the user keeps their hidden service address private it should be ok, but maybe it would be better to have this disabled by default.

If a user is using one of the few applications that requires RPC access, they can just flick a toggle (maybe with a warning of the dangers), to enable it.

mayankchhabra:
Ah yes, you're right. Sorry my bad, I confused the RPC port number with the P2P port number.

Regarding keeping the RPC address private, yeah we can hide it by-default and add a warning. But RPC hidden service URL alone won't be useful for the user without the RPC username and password. Maybe that should be handled by `umbrel-manager`.

lukechilds:
>Regarding keeping the RPC address private, yeah we can hide it by-default and add a warning.

I think the actual Tor hidden service should be disabled until they enable it, not just hidden from the UI.

>But RPC hidden service URL alone won't be useful for the user without the RPC username and password. Maybe that should be handled by umbrel-manager.

Yeah definitely, we should just give them a full connection string like `http://user:pass@hiddenservice.onion:port`.

mayankchhabra:
> I think the actual Tor hidden service should be disabled until they enable it, not just hidden from the UI.

I understand your point, but in your opinion what is the biggest risk of accidentally leaking RPC access? 

From my understanding, as long as we're not using the `bitcoind` wallet, leaking RPC connection info is far less risky than leaking your dashboard's onion address and password.

lukechilds:
>I understand your point, but in your opinion what is the biggest risk of accidentally leaking RPC access?

There's not a specific attack I have in mind, it's more that I just feel a little bit uncomfortable about exposing it since I don't think it's been explicitly designed to be publicly exposed.

The assumed use case is that it's only accessible inside a trusted(ish) environment (the local network). There could potentially be unknown vulnerabilities, for example RCE, that wouldn't have been strictly tested for like you would with a public facing API.

If people want to expose it, that's ok and we should support it, I just don't think we should be doing it by default.

> From my understanding, as long as we're not using the bitcoind wallet, leaking RPC connection info is far less risky than leaking your dashboard's onion address and password.

If an attacker gets RPC access they have complete control of your node. Even if you don't have a wallet, the attacker could remove genuine peers and add peers they control to perform a network partition attack. The attacker would then have control over what your full node sees. They can censor transactions from you and send you Bitcoins that don't exist.

mayankchhabra:
Fair enough. Perhaps we should have `umbrel-manager` update `torrc` to create the hidden service on user request. 

But on a second thought, I agree with your initial skepticism about even having this info available at all, or at least in the public-beta. Plus there are not many consumer-friendly use-cases right now that require access to `bitcoind`'s RPC.

If we all agree to that, we can reconsider implementing this after the public-beta.

lukechilds:
I agree, I don't think this is important for the beta.

mayankchhabra:
I realized that we actually meant to show the P2P onion address and not the RPC address, since the goal is to allow people to connect their Umbrel's Bitcoin Core node to their wallet, eg. Wasabi, which just needs to add it as a peer using the P2P hidden service address. 

Updated https://github.com/getumbrel/umbrel-dashboard/issues/125 to reflect the same, and closing this now. 

@nolim1t feel free to take this up if you'd like too. We'd probably need a new endpoint in `umbrel-middleware`.

PR #38, opened 2020-07-19
Add in watchtower support so we can utilize it later

nolim1t:
Add in watchtower support so we can utilize it later in having each node watch out for each other and have a secure network (see [watchtower notes](https://github.com/lightningnetwork/lnd/blob/master/docs/watchtower.md))

### DIFF
--- a/lnd/lnd.conf
+++ b/lnd/lnd.conf
@@ -53,6 +53,15 @@ neutrino.addpeer=mainnet2-btcd.zaphq.io
 ;neutrino.addpeer=testnet2-btcd.zaphq.io
 
 
+; activate watchtowers so we have a strong watchtower network
+[watchtower]
+watchtower.active=1
+
+; activate watchtower client so we can get get other servers
+; to make sure noone steals our money
+[wtclient]
+wtclient.active=1
+
 [tor]
 tor.active=1
 tor.control=10.11.5.1:29051


mayankchhabra:
Good line of thought, @nolim1t. Having Umbrel users added to each other's Watchtowers would be cool, but here's why I don't think it's a good idea right now:

- This PR enables watchtower server and client, but none of them will actually be used because no clients are added to the any of the watchtower servers. 

- PRs should add features in full. Unimplemented features can make the codebase confusing and harder to maintain. 

- We do not specifically need to enable watchtower right now. They can be enabled later via an OTA update once they're implemented in full.

- The [task-list for Umbrel Public Beta](https://github.com/orgs/getumbrel/projects/1) is already too tight and it'd be great to finish those tasks first, else we can end up delaying our launch.

- Long term tasks like this one should be added to the [medium/long-term idea board](https://github.com/orgs/getumbrel/projects/2) first, so we can brainstorm on them when the time comes and plan their implementation.

nolim1t:
Its complete from this repos point of view. It enables it for later and allows people to enable it from the command line.

For the users to access it through the UI, it has to be done separately as a middleware and dashboard task.

This is the first step

mayankchhabra:
I understand the PoV, but if people still have to update `lnd.conf` from the command line to use watchtowers, they can themselves enable the watchtowers too.

nolim1t:
I'm just laying the groundwork all ready. It doesn't impact anything

nolim1t:
Part of #212

mayankchhabra:
Closing this due to inactivity.

PR #52, opened 2020-07-25
Start and stop Umbrel with native scripts

mayankchhabra:
This PR implements native start and stop scripts for Umbrel, allowing us to update the start/stop logic on the go via OTA updates, and create/access/configure a custom runtime environment.


start
```diff
@@ -27,7 +27,8 @@ echo "Starting karen..."
 echo
 ./karen &
 KAREN_PID=$!
-echo $! > ./karen.pid
+KAREN_PGID=$(ps opgid= "$KAREN_PID")
```
lukechilds, line 30: Do we want to do this?

Not tested but doesn't this mean if we do a shutdown then we run `./stop` to kill karen and all it's child processes including the shutdown child process and then prevent the shutdown?

Not sure if shutdown works like that. Maybe at that point the shutdown process is already being handled by the kernel so killing `shutdown` doesn't do anything, I'm not sure. Needs testing though.

start
```diff
@@ -0,0 +1,35 @@
...
+
+check_dependencies fswatch readlink dirname docker docker-compose rsync jq wget git
+
+echo "Starting Umbrel..."
+echo
+echo "Starting karen..."
+echo
+./karen &
+KAREN_PID=$!
+echo $! > ./karen.pid
```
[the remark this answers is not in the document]
mayankchhabra: Yeah, also using PGID now to kill all child processes

start
mayankchhabra: Ah, you're right. The shutdown doesn't work if karen and her children are killed lol.

The goal was to not spawn multiple instances of karen. And I think you're right, the approach of killing karen isn't right and will break not only shutdown and reboot, but also the OTA updates. 

We should probably add a check in karen itself to not run if it's already running.

lukechilds: >We should probably add a check inside karen itself to not run if it's already running.

Yeah definitely, that sounds like a good approach.

We can write a pidfile inside `karen` to the events dir it's watching, then remove it when the script exits with `trap`.

mayankchhabra: Yes, but then we'll also have to check and remove that file on boot as it can exist after a power failure/crash. Perhaps we can just check against `ps -A | grep karen` (inside karen, before starting up) for now...

lukechilds: That's true, `ps -A | grep karen` won't work because it's actually `bash ./karen` not `karen`. You can do `ps -Af | grep karen` but there is a higher potential for false positives.

e.g `someprocess --that-is-definitely-no-karen` will return as a match for `ps -Af | grep karen`.

I think it should be fine though.

mayankchhabra: Ah okay, understood. Sounds good, opened a separate issue for it #57

scripts/start
```diff
@@ -0,0 +1,43 @@
...
+check_dependencies () {
+  for cmd in "$@"; do
+    if ! command -v $cmd >/dev/null 2>&1; then
+      echo "This script requires \"${cmd}\" to be installed"
+      exit 1
+    fi
+  done
+}
+
+check_dependencies fswatch readlink dirname docker docker-compose rsync jq wget git
```
lukechilds, line 22: Why are these all added here?

For example it's not clear to me where or why `jq` is required.

And if it's removed from whichever script requires it we might forget to remove it here.

If they need to be added here because the other scripts will error later rather than at start time maybe we could add some comments here explaining which files require which deps.

scripts/status
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Get Umbrel Status
+
+if [[ $UID != 0 ]]; then
+    echo "Umbrel's status must be fetched as root"
+    echo "Please re-run this script as"
+    echo "  sudo ./scripts/status"
+    exit 1
```
lukechilds, line 10: Is root required for this?

The current user should be added to the docker group so I think it should work without root.

And if they aren't in the docker group then docker should give them it's won warning.

scripts/start
```diff
@@ -0,0 +1,43 @@
...
+
+echo "Starting Umbrel..."
+echo
+
+UMBREL_ROOT="$(dirname $(readlink -f ${BASH_SOURCE[0]}))/.."
```
lukechilds, line 27: Missed some quotes here.

```suggestion
UMBREL_ROOT="$(dirname $(readlink -f "${BASH_SOURCE[0]}"))/.."
```

Don't worry, that was a tricky one...

scripts/stop
```diff
@@ -0,0 +1,24 @@
...
+# Stop Umbrel
+
+if [[ $UID != 0 ]]; then
+    echo "Umbrel must be stopped as root"
+    echo "Please re-run this script as"
+    echo "  sudo ./scripts/stop"
+    exit 1
+fi
+
+UMBREL_ROOT="$(dirname $(readlink -f ${BASH_SOURCE[0]}))/.."
```
lukechilds, line 13: Missing quotes.

```suggestion
UMBREL_ROOT="$(dirname $(readlink -f "${BASH_SOURCE[0]}"))/.."
```

scripts/start
mayankchhabra: Ah yes, comments would be nice. `jq` is used by the OTA update trigger by the way. 

The reason I added them here was to make sure something doesn't go wrong **after** the user is up and running Umbrel, because for instance, let's say `jq` isn't installed on the host and the OTA update fails, there's no way the user would be able to know what went wrong.

scripts/status
mayankchhabra: Ah okay, I just added it for consistency among all the three scripts.

mayankchhabra: Will remove it, since it can be added later if root access is actually required.

Issue #36, opened 2020-07-18
Existing Umbrel drive doesn't work with a clean install of Umbrel OS


mayankchhabra:
Our [partitioning utility](https://github.com/getumbrel/umbrel/blob/master/contrib/partitioner/partitioner.py) needs to be fixed to support existing Umbrel drives, so even if the user re-flashes Umbrel OS on their SD card, they can continue where they left-off.

nolim1t:
what happens.. it should already not wipe anything unless specified

mayankchhabra:
Yeah you're right, it doesn't wipe the data. But there's still some data loss. Eg. `docker-compose.yml` is stored on the SD card, but when the SD card is re-flashed, it doesn't work with the existing drive.

nolim1t:
I think we'll need to copy the docker-compose file to the drive too

mayankchhabra:
Yeah, but we should perhaps redo some of the logic in the partitioner. Eg. if a user connects a drive with an existing `bitcoin` directory, it still wouldn't work with Umbrel because the RPC credentials in that `bitcoin` directory wouldn't match the ones we generate. 

IMO we should completely nuke the drive in all cases, as long as there is no `.umbrel` file in the drive's root.

nolim1t:
Don't want to nuke the drive by default... that would be mean to make people sync the blockchain again. Especially if someone pays for bandwidth.

mayankchhabra:
I understand, but then we need to a create a foolproof partitioning logic.

It can be something like this (on the first boot):

- Check if the drive is an existing Umbrel drive by looking for a `.umbrel` file in drive's root.
- If yes, don't touch anything on the drive and just setup all the required symlinks.
- If no `.umbrel` file is present, check if it has `bitcoin/blocks` and `bitcoin/chainstate` directories. 
- If not, nuke it.
- If yes, nuke everything other than `bitcoin/blocks` and `bitcoin/chainstate` (as we only need to preserve blockchain data and nothing else).

nolim1t:
Maybe better to check for other data files too and also try to preserve docker-compose

nolim1t:
But I'd rather not put too much install specific stuff inside a compose file to be honest and use config files for each container  instead. 

That way containers are more configurable and we don't have to restart the whole docker-compose when something changes.

mayankchhabra:
> Maybe better to check for other data files too and also try to preserve docker-compose

[...]

PR #54, opened 2020-07-26
Add docker-compose file to the drive

nolim1t:
This Closes https://github.com/getumbrel/umbrel/issues/36

We will run docker-compose.yml as a symlink to prevent inconsistencies with generated data which occur if the SD card is reflashed


mayankchhabra:
LGTM

PR #72, opened 2020-07-31
Move karen to scripts/karen

mayankchhabra:


mayankchhabra:
This requires some rewiring on karen's end as the triggers mess up the paths. Not needed on priority so we can patch it in an OTA update.

PR #73, opened 2020-07-31
RAM fixes

AaronDewes:
Replaces https://github.com/getumbrel/umbrel-os/pull/75

### DIFF
--- a/bitcoin/bitcoin.conf
+++ b/bitcoin/bitcoin.conf
@@ -14,7 +14,7 @@ rest=1
 blocksonly=1
 maxconnections=8
 dbcache=1000
-maxmempool=512
+maxmempool=300
 maxuploadtarget=5000
 
 # uncomment test if testnet=1 is activated
--- a/scripts/configure
+++ b/scripts/configure
@@ -97,6 +97,10 @@ if [ -z $TESTNET ] && [ -z $REGTEST ]; then
     sed -i 's/RPCPORT/8332/g; ' docker-compose.yml
 fi
 
+echo "Setting dbcache size"
+DBCACHE_SIZE=$(awk '/MemTotal/{printf "%d\n", ($2/2^10 * 0.5) - 300}' /proc/meminfo)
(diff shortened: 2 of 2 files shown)


mayankchhabra:
LGTM. Thanks, @AaronDewes! 

Can you please move the logic from `scripts/start` to `scripts/configure` instead as it's a one-time configuration.

scripts/start
```diff
@@ -46,6 +46,10 @@ export DEVICE_HOST="http://"${DEVICE_IP:-"$DEVICE_HOSTNAME".local}""
 
 cd "$UMBREL_ROOT"
 
+echo "Setting dbcache size"
+DBCACHE_SIZE=$(awk '/MemTotal/{printf "%d\n", $2/2^10 * 0.5}'  /proc/meminfo)
+sed -i -e "s/dbcache=1000/dbcache=$FIXED/g" bitcoin/bitcoin.conf
```
lukechilds, line 51: Looks like you had the wrong variable name left in from development?

```suggestion
sed -i -e "s/dbcache=1000/dbcache=$DBCACHE_SIZE/g" bitcoin/bitcoin.conf
```

scripts/start
```diff
@@ -46,6 +46,10 @@ export DEVICE_HOST="http://"${DEVICE_IP:-"$DEVICE_HOSTNAME".local}""
 
 cd "$UMBREL_ROOT"
 
+echo "Setting dbcache size"
+DBCACHE_SIZE=$(awk '/MemTotal/{printf "%d\n", $2/2^10 * 0.5}'  /proc/meminfo)
```
lukechilds, line 50: This sets `DBCACHE_SIZE` to 50% of RAM. It should be 50%-300 to account for the 300MB allocated to `maxmempool.`

```suggestion
DBCACHE_SIZE=$(awk '/MemTotal/{printf "%d\n", ($2/2^10 * 0.5) - 300}' /proc/meminfo)
```

lukechilds:
Thanks for taking this on @AaronDewes, some tiny changes needed.

PR #79, opened 2020-08-01
IN TESTING: Get rid of any references to home directory so this can run from anywhere in the filesystem

nolim1t:
Get rid of any references to home directory in partitioner so this can run from anywhere in the filesystem and not just /home/umbrel.


mayankchhabra:
Hey @nolim1t, I think this has been taken care of by #93.

PR #91, opened 2020-08-03
Use github action built (and GPG verified) tor service

nolim1t:
Have a tor container based on debian-slim which is GPG verified and built for cross platform (from the official sources)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,7 +12,7 @@ x-utility: &default-utility
 services:
         tor:
                 container_name: tor
-                image: getumbrel/tor:v0.4.1.9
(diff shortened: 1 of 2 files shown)


nolim1t:
This looks good

mayankchhabra:
It'd be really great if we could upgrade and use https://github.com/getumbrel/docker-tor

nolim1t:
I'm not the maintainer of that project

mayankchhabra:
But you can make a PR?

mayankchhabra:
Hey, @nolim1t closing this due to inactivity. Feel free to make a PR on https://github.com/getumbrel/docker-tor with your upgrades if you'd like to.

PR #98, opened 2020-08-07
Bump lnd version

nolim1t:
Bump LND version to ```lncm/lnd:v0.10.4-experimental```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -56,7 +56,7 @@ services:
                         ipv4_address: 10.11.1.1
         lnd:
                 container_name: lnd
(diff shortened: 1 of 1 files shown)


mayankchhabra:
Would be great if we can shift to standard "non-experimental" lnd images. How important in your opinion is `routing.assumechanvalid=1`?

nolim1t:
Better UX..

mayankchhabra:
How, can you explain? Genuinely want to understand what it does.

nolim1t:
speeds up transactions for nearby peers

PR #100, opened 2020-08-08
Replace init script with systemd services

mayankchhabra:
- `systemd` is the facto init system for Linux.
- Greater dependency control. For instance, we can now guarantee that Umbrel will **NOT** start if the external storage mounter fails for whatever reason.
- Faster boot time as we can run services in parallel without blocking the boot process.
- Choice of logging. For instance, we can now log `umbrel-startup.service` and `umbrel-external-storage.service` to `syslog`, and `umbrel-connection-details` service to `tty`.
- Overall more verbose, clean and straightforward than using `rc.local` + `initd`.


scripts/umbrel-os/boot
```diff
@@ -3,49 +3,50 @@
...
+  echo "=========== hello world! ============="
+  echo "======================================"
+  echo 
+  echo "Configuring Umbrel..."
   echo
-  if [ -f "$UMBREL_ROOT"/scripts/configure ]; then
+  if [ -f "${UMBREL_ROOT}/scripts/configure" ]; then
     cd "$UMBREL_ROOT"
-    ./scripts/configure || exit 1
+    NETWORK=regtest ./scripts/configure || exit 1
```
mayankchhabra, line 40: Change this to `mainnet` before merge

Issue #108, opened 2020-08-09
Load newer install of Umbrel from SD card to SSD


mayankchhabra:
So that if someone re-flashes Umbrel OS that has the latest version of Umbrel, they can run that version instead of the older version on their SSD.

mayankchhabra:
Closed by #134

PR #115, opened 2020-08-10
Install OTA update on the SD card too

### LINKED ISSUE #107: Install OTA update on both SD Card and SSD

mayankchhabra:
Since `/home/umbrel/umbrel` is bind-mounted on Umbrel OS, the OTA update currently updates the Umbrel installation only on the external storage. That means updating the `/scripts/umbrel-os/boot` script (or any other script that's used by the system **before** the external storage is mounted) has no effect.

This PR updates the OTA functionality to update the SD card installation along with the external storage installation. 

Closes #107


lukechilds:
Instead of unmounting the bind mount, accessing what's underneath, and then remounting it, you can just bind mount the root filesystem somewhere else and access it directly from that mount point.

`mount --bind` is non-recursive, so you will be able to access the original filesystem, without the `$UMBREL_ROOT` bind mount applied.

e.g:

```shell
mkdir /original-root
mount --bind / /original-root
# Access SDcard install at: "/original-root${UMBREL_ROOT}"
```

mayankchhabra:
Wow, that sure was one ingenious idea @lukechilds! Made things incredibly simpler and safer.

scripts/update/01-run.sh
```diff
@@ -4,6 +4,9 @@ set -euo pipefail
 RELEASE=$1
 UMBREL_ROOT=$2
 
+# Only used on Umbrel OS
+SD_CARD_UMBREL_ROOT="/sd-root/home/umbrel/umbrel"
```
lukechilds, line 8: Shouldn't we use `/sd-root${UMBREL_ROOT}` here?

mayankchhabra: Ah, yes absolutely. Updated!

scripts/update/01-run.sh
```diff
@@ -12,6 +15,28 @@ echo "=========== Stage: Install ============"
...
+
+    # Cleanup any tmp dir from unclean install run
+    [[ -d "/home/umbrel/.umbrel-${RELEASE}" ]] && rm -rf "/home/umbrel/.umbrel-${RELEASE}"
+
+    # Copy from external storage to SD card
+    mkdir -p "/home/umbrel/.umbrel-$RELEASE"
+    cp  --recursive \
+        --archive \
+        --no-target-directory \
+        "$UMBREL_ROOT/.umbrel-$RELEASE" "/home/umbrel/.umbrel-$RELEASE"
```
lukechilds, line 37: Why do we need to copy to this temp directory? "/home/umbrel/.umbrel-$RELEASE"

Can't we just `rsync` directly from `"$UMBREL_ROOT/.umbrel-$RELEASE"` to `$SD_CARD_UMBREL_ROOT `?

mayankchhabra: Ahhh, yes. You're right, this doesn't make any sense now. It was there before because of the mount/unmount logic (that's why an intermediary dir was needed), but since we don't have that problem now, we can just directly copy to the SD card root.

mayankchhabra: Fixed

mayankchhabra:
Perfect, thanks for your reviews!

No worries regarding testing... I tested it quite a few times, but nonetheless we'll test everything again via an OTA update after #108 is fixed!

PR #125, opened 2020-08-12
Include mount check in mount script

lukechilds:
If we call an external script (scripts/umbrel-os/external-storage/is-mounted) to do the check, if the drive is an old version where the script doesn't exist at that location, the check will fail, even if external storage was mounted successfully.

Including the check directly in the script means we know exactly what will execute, regardless of what's going on with the filesystem.

I originally wrote the mount check as an external script so it could be re-used by other things, but we're not doing that yet so I've deleted it for now.


mayankchhabra:
Sounds good! 👌

PR #134, opened 2020-08-13
Update external Umbrel installation when booting from a newer Umbrel OS version.

lukechilds:
This is pretty much completely untested so far, just opening a draft PR so people can play around with it.

To see the OTA update logic applied at the filesystem level you can run:

```
sudo scripts/update/update --path /a/different/umbrel/install/to/update/to
```

To see the semver checks and how this will be initiated run:

```
sudo scripts/umbrel-os/external-storage/update-from-sdcard
```

I need to create a new service that runs this directly after the mount logic and before anything else.

I also need to test it quite a bit and also check I haven't broken anything with OTA updates.


scripts/umbrel-os/external-storage/update-from-sdcard
```diff
@@ -0,0 +1,48 @@
...
+SD_MOUNT_POINT="/sd-root"
+SD_UMBREL_ROOT="${SD_MOUNT_POINT}${UMBREL_ROOT}"
+
+check_root () {
+  if [[ $UID != 0 ]]; then
+    echo "This script must be run as root"
+    exit 1
+  fi
+}
+
```
mayankchhabra, line 15: Should we add a dependency check here for `jq`?

lukechilds: Ah yes, good spot.

mayankchhabra:
Wow, that's a super-slick implementation, @lukechilds! The code looks great.

Looking forward to testing it with the Umbrel OS image being built on your fork.

Re: 
> I need to create a new service that runs this directly after the mount logic and before anything else.

Do you think creating a new service is necessary? I think we can club it as a part of the entire "external storage" logic. So instead of a new service, we can add 
```
ExecStartPost=/home/umbrel/umbrel/scripts/umbrel-os/external-storage/update-from-sdcard
```
To the `umbrel-external-storage.service`
https://github.com/getumbrel/umbrel/blob/6f634bb296f949aac3d7dc1f545c793a1016e465/scripts/umbrel-os/services/umbrel-external-storage.service#L10

It will also be respected by the `Before/After` constraints set by other services.

lukechilds:
Nice, didn't know about `ExecStartPost`, yeah that seems a lot cleaner.

scripts/umbrel-os/services/umbrel-external-storage.service
```diff
@@ -8,6 +8,7 @@ Description=External Storage Mounter
 Type=oneshot
 Restart=no
 ExecStart=/home/umbrel/umbrel/scripts/umbrel-os/external-storage/mount
+ExecStartPost=/home/umbrel/umbrel/scripts/umbrel-os/external-storage/update-from-sdcard
```
mayankchhabra, line 11: Do you think it's a good idea to increase the default timeout time from the default 90 seconds to something that's much more generous, like 45 minutes? The reason being that if there's a version mismatch and `update-from-sdcard` triggers the OTA update process, it could take a long while to download and extract new Docker images, especially on a slow network.

```suggestion
ExecStartPost=/home/umbrel/umbrel/scripts/umbrel-os/external-storage/update-from-sdcard
TimeoutStartSec=45min
```

mayankchhabra:
Ran some tests and everything worked flawlessly! Great job, @lukechilds! 👏

Test 1. SD Card Umbrel version = SSD Umbrel version
```
...
umbrel external storage mounter[378]: Mount script completed successfully!
umbrel external storage mounter[740]: Checking if SD card Umbrel is newer than external storage...
umbrel external storage mounter[740]: No, SD version is not newer, exiting.
```

Test 2. SD Card Umbrel version > SSD Umbrel version, but doesn't fulfil the required condition for update
```
...
umbrel external storage mounter[349]: Mount script completed successfully!
umbrel external storage mounter[526]: Checking if SD card Umbrel is newer than external storage...
umbrel external storage mounter[526]: Yes, SD version is newer.
umbrel external storage mounter[526]: Checking if the external storage version "0.1.6" satisfies update requirement "=0.1.7
umbrel external storage mounter[526]: No, we can't do an automatic update, exiting.
```

Test 3. SD Card Umbrel version < SSD Card Umbrel version

```
...
umbrel external storage mounter[359]: Mount script completed successfully!
umbrel external storage mounter[522]: Checking if SD card Umbrel is newer than external storage...
umbrel external storage mounter[522]: No, SD version is not newer, exiting.
```

Test 4. SD Card Umbrel > SSD Umbrel, and fulfils the required condition for update

```
...
umbrel external storage mounter[364]: Mount script completed successfully!
umbrel external storage mounter[524]: Checking if SD card Umbrel is newer than external storage...
umbrel external storage mounter[524]: Yes, SD version is newer.
umbrel external storage mounter[524]: Checking if the external storage version "0.1.8" satisfies update requirement ">=0.1.
umbrel external storage mounter[524]: Yes, it does, attempting an automatic update...
...
umbrel external storage mounter[524]: Successfully installed Umbrel 0.1.9
umbrel external storage mounter[524]: == End Update Script 03-run.sh ==
umbrel external storage mounter[524]: Deleting cloned repository
umbrel external storage mounter[524]: Removing lock
umbrel systemd[1]: Started External Storage Mounter.
```

PR #136, opened 2020-08-14
Improve actions

AaronDewes:
The old check tried to build from the PR branch for merge commits. If the PR branch is deleted immediately after the merge, the check still tries to clone the PR repo, and fails.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -11,7 +11,6 @@ jobs:
     runs-on: ubuntu-20.04
 
     steps:
-    - uses: actions/checkout@v1
     - name: Run apt update
       run: sudo apt update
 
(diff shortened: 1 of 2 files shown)


.github/workflows/push.yml
```diff
@@ -0,0 +1,70 @@
+name: Build OS image of current master
+
+on:
+  push:
+    branches:
+      - master
```
AaronDewes, line 6: If this is only `on: push`, it also starts on PR.

AaronDewes:
As you can see, https://github.com/getumbrel/umbrel/runs/987433277#step:11:15 thinks it is building a PR, but it is building a merge commit in the main umbrel repo, which merged a now deleted branch, causing the action to fail. This PR fixes that issue.

mayankchhabra:
Awesome! Thanks for the fix, @AaronDewes

PR #139, opened 2020-08-15
Reformat external storage if it isn't EXT4

### LINKED ISSUE #103: External storage mounter fails to mount exFAT device
When checking for `$UMBREL_EXTERNAL_ROOT/.umbrel`, we should first check if the partition is `ext4`, and if not, nuke it.

lukechilds:
Resolves #103

### DIFF
--- a/scripts/umbrel-os/external-storage/mount
+++ b/scripts/umbrel-os/external-storage/mount
@@ -54,6 +54,11 @@ get_block_device_model () {
   echo "$(echo $vendor) $(echo $model)"
 }
 
+is_partition_ext4 () {
+  partition_path="${1}"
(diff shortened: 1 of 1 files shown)


mayankchhabra:
Awesome, let's merge this directly since you would've tested it already.

lukechilds:
Haven't tested it quite yet, gimme 5 mins.

lukechilds:
Also testing stuff like this is so quick now with:

```
sudo mkdir /lukechilds-umbrel
cd /lukechilds-umbrel/
sudo chown umbrel:umbrel $PWD
git clone https://github.com/lukechilds/umbrel.git .
git checkout check-external-filesystem
sudo ~/umbrel/scripts/update/update --path /lukechilds-umbrel/
reboot
```

lukechilds:
Tested and working.

PR #142, opened 2020-08-15
Remove access control headers

lukechilds:
I'm not sure why these access control headers are added but they don't seem to be required and everything appears to still work when they're removed.

They seem potentially dangerous, especially `Access-Control-Allow-Origin *` which would allow a malicious website to make requests directly to Umbrel services on the user's local network and read the responses.

The browser defaults to a very strict CORS policy and these headers loosen it significantly. I don't think we need to loosen the rules because all the Umbrel services are always accessed from the same origin.

@mayankchhabra @nolim1t let me know if there was a reason these rules were added and if they are required for something.

### DIFF
--- a/nginx/conf.d/default.conf
+++ b/nginx/conf.d/default.conf
@@ -4,17 +4,14 @@ server {
     server_name _;
 
     location /api/ {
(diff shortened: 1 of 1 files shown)


mayankchhabra:
LGTM! What a good find, @lukechilds.

Issue #96, opened 2020-07-31
Can't open channels, mempool rejects low fee transaction


mayankchhabra:
Posted by @shawnyeager on our Telegram:

![image](https://user-images.githubusercontent.com/10330103/88997347-15ae4700-d30d-11ea-8649-02ff4d76cd1e.png)

nolim1t:
moved back to umbrel project as its just a config

PR #152, opened 2020-08-16
Clean up Bitcoin Core config

lukechilds:
I've removed the following:

```
server=1
```
Redundant, it's on by default

```
rest=1
```
As far as I can see we're not using REST for anything.

```
maxorphantx=10
```
Not sure why we've set this, the default is 100 which is not much different. I don't think there's a valid reason to set this to something non-standard.

```
minrelaytxfee=0.00000001
```
There's no benefit to this, it's actually a hindrance since it means our mempool will get filled up with low fee transactions that will probably never confirm. Our mempool will be further out of sync with most other nodes. We should use the default relay fee to maintain a closer mempool state to the rest of the network.

```
prune=0
```
Redundant, this is the default value. If we need to add pruning based on available storage in the future we can just add it dynamically.

```
# Automatically uncommented if Umbrel
# is configured for testnet
#testnet=1
#[test]

# Automatically uncommented if Umbrel
# is configured for regtest
#regtest=1
#[regtest]
```
I've removed the commented out testnet/regtest sections. These are now just dynamically added to the config by the configure script when the appropriate network has been selected.

I've changed the Tor config from:

```
onion=10.11.5.1:29050
torcontrol=10.11.5.1:29051
torpassword=<password>
```

to:

```
proxy=10.11.5.1:29050
listen=1
bind=10.11.1.1
```

This was previously creating and reporting a hidden service that was unreachable. It was also only using the Tor proxy to access Tor nodes, but clearnet nodes were accessed over clearnet.

Now all outgoing network communication happens over Tor. Regardless of whether the node we are connecting to is running on clearnet or Tor, we always route the connection through Tor so our IP is not exposed.

No hidden service is created or reported by Bitcoin core. However we are now manually creating a hidden service in the Tor config and proxying it to the Bitcoin Core P2P port. You can view the address at `tor/data/bitcoin-p2p/hostname`. Bitcoin Core is unaware of this hidden service, so doesn't broadcast it, however it will still respond to connections to the hidden service. This means you can still manually set up Umbrel as a trusted node in wallets such as Wasabi, Samourai or any SPV wallet.

Please let me know if you think any of the configuration options I've changed/removed were needed for a specific reason.

### DIFF
--- a/scripts/configure
+++ b/scripts/configure
@@ -107,10 +107,8 @@ if [ "$BITCOIN_NETWORK" == "testnet" ]; then
   # Set testnet ports
   BITCOIN_RPC_PORT=18332
   BITCOIN_P2P_PORT=18333
-  # Uncomment "test" block
-  sed -i "s/\#\[test\]/\[test\]/g;" "$BITCOIN_CONF_FILE"
-  # Enable testnet
-  sed -i "s/\#testnet=1/testnet=1/g" "$BITCOIN_CONF_FILE"
+  # Switch Bitcoin Core to testnet
+  sed -i '1s/^/testnet=1\n[test]\n\n/' "$BITCOIN_CONF_FILE"
   # Switch LND to testnet
   sed -i "s/bitcoin.mainnet=1/bitcoin.testnet=1/g;" "$LND_CONF_FILE"
   # Uncomment testnet neutrino block and peers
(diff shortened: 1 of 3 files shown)


scripts/configure
```diff
@@ -107,10 +107,8 @@ if [ "$BITCOIN_NETWORK" == "testnet" ]; then
   # Set testnet ports
   BITCOIN_RPC_PORT=18332
   BITCOIN_P2P_PORT=18333
-  # Uncomment "test" block
-  sed -i "s/\#\[test\]/\[test\]/g;" "$BITCOIN_CONF_FILE"
-  # Enable testnet
-  sed -i "s/\#testnet=1/testnet=1/g" "$BITCOIN_CONF_FILE"
+  # Switch Bitcoin Core to testnet
+  sed -i '1s/^/testnet=1\n[test]\n\n/' "$BITCOIN_CONF_FILE"
```
mayankchhabra, line 111: Love this 😍

templates/bitcoin-sample.conf
```diff
@@ -29,9 +14,6 @@ rpcauth=<rpcauth>
 
 # Optimizations
 dbcache=<size>
-maxconnections=40
 maxmempool=300
-maxorphantx=10
+maxconnections=40
 maxuploadtarget=1000
-minrelaytxfee=0.00000001
```
mayankchhabra, line 36: Ah, this is actually required for https://github.com/getumbrel/umbrel/issues/96

lukechilds: According to the docs:

```
  -minrelaytxfee=<amt>
       Fees (in BTC/kB) smaller than this are considered zero fee for relaying,
       mining and transaction creation (default: 0.00001)
```

So by setting minrelaytxfee=0.00000001 we're saying we relay 1 sat/kb not 1 sat/b.

The default value of 0.00001, is 0.00001 BTC/kb, which is 1000 sat/kb, which is 1 sat / byte.

I don't think we should relay <1 sat/b transactions since that could lead to users broadcasting transactions that never confirm.

As discussed on Keybase, I can remove that change for now until we have time to investigate the fee issue further.

PR #161, opened 2020-08-18
Add CLI binary proxies

lukechilds:
This PR creates a new `$UMBREL_ROOT/bin` directory that should be added to the users `$PATH`.

We can do this automatically on Umbrel OS but users installing themselves can also add this to their PATH manually.

Currently there are only two binaries, `lncli` and `bitcoin-cli`, which just pass all arguments directly through to the equivalent binary inside the docker container and then return the correctly formatted output.

This allows you to use  globally available `lncli` and `bitcoin-cli` binaries and pipe around their output as if they were installed on the host sytem.

Example:

```shell
umbrel@umbrel:~ $ export PATH="$PATH:$HOME/umbrel/bin"
umbrel@umbrel:~ $ bitcoin-cli --version
Bitcoin Core RPC client version v0.20.1.0-g7ff64311bee570874c4f0dfa18f518552188df08
umbrel@umbrel:~ $ lncli --version
lncli version 0.10.4-beta commit=v0.10.4-beta
```

It makes debugging and development a lot simpler and I think it's worth including directly in Umbrel OS to make it easier when we're helping user's to debug things too.


mayankchhabra:
Looks great, @lukechilds! Definitely simplifies debugging quite a lot.

It seems as though `bin/lncli` isn't executable. Can you double check the permissions once?

lukechilds:
Oops, good spot, fixed.

mayankchhabra (approved):

lukechilds:
@mayankchhabra do you think we should document this as an extra step in the installation instructions in the readme?

Or maybe it's fine for us to just leave this for people to find or we can tell people to add it to their path if they're debugging. We don't really want to encourage manual CLI use.

mayankchhabra:
> Or maybe it's fine for us to just leave this for people to find or we can tell people to add it to their path if they're debugging. We don't really want to encourage manual CLI use.

Totally agree! It's mostly for us to help them debug.

PR #162, opened 2020-08-19
Remove git dependency check from scripts

lukechilds:
Correct me if I'm wrong but I don't think the git dependency is needed anymore since updates are now performed via `curl`.

`grep`ing the entire code base shows these dependency checks are the only occurrences of `git`.

### DIFF
--- a/scripts/start
+++ b/scripts/start
@@ -20,10 +20,10 @@ check_dependencies () {
 }
 
 # Check system's dependencies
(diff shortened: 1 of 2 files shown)


mayankchhabra:
Yeah, looks like it's not being used anywhere. Good spot!

mayankchhabra (approved):

PR #179, opened 2020-08-24
Remove default maxmempool value

mayankchhabra:
Since the [default value](https://github.com/bitcoin/bitcoin/blob/0d71395848bbc2941862e7a0644f47212ec02e87/doc/reduce-memory.md#memory-pool) of maxmempool is 300mb, we probably don't need to specify it in the config. @lukechilds can you please confirm?

### DIFF
--- a/templates/bitcoin-sample.conf
+++ b/templates/bitcoin-sample.conf
@@ -14,6 +14,5 @@ rpcauth=<rpcauth>
 
 # Optimizations
 dbcache=<size>
-maxmempool=300
 maxconnections=40
 maxuploadtarget=1000


lukechilds:
It's correct that the default value is 300, however I explicitly set it in the config since we rely on that value for the Bitcoin Core memory usage.

We set `dbcache` to `($total_ram/2)-300` because we know `maxmempool` is 300. So we aim to let Bitcoin Core utilise 50% of the available system memory.

We could remove the explicit `maxmempool` line from the config, but that means if a future version of Bitcoin Core increases the default mempool limit then our memory calculations will be off.

mayankchhabra:
Ahhh okay, that makes sense. I didn't realize that 300 in dbcache calculation came from maxmempool.

lukechilds:
Yeah the other reason those two values are related is that `dbcache` is allowed to overflow into any unused `maxmempool` space.

mayankchhabra:
Nicee!

PR #180, opened 2020-08-25
Create 4GB swapfile on external storage

### LINKED ISSUE #104: Automatically configure and use swap
Create a 2GB swapfile on the SSD after the mount. This is crucial for low-RAM devices such as RPi 3 and RPi 4 /w 2GB RAM or else bitcoin crashes during the sync.

lukechilds:
Resolves #104 

Bind mount `/mnt/data/swap` on external storage to `/swap` and create 4GB swapfile at `/swap/swapfile`.

### DIFF
--- a/scripts/umbrel-os/external-storage/mount
+++ b/scripts/umbrel-os/external-storage/mount
@@ -170,8 +170,10 @@ main () {
   mount --bind "${EXTERNAL_UMBREL_ROOT}" "${UMBREL_ROOT}"
 
   echo "Bind mounting external storage to /swap"
+  systemctl stop dphys-swapfile
   mkdir -p "${MOUNT_POINT}/swap" "/swap"
   mount --bind "${MOUNT_POINT}/swap" "/swap"
+  systemctl start dphys-swapfile
 
   echo "Bind mounting SD card root at /sd-card..."
   [[ ! -d "/sd-root" ]] && mkdir -p "/sd-root"
--- a/scripts/umbrel-os/external-storage/mount
+++ b/scripts/umbrel-os/external-storage/mount
@@ -16,6 +16,8 @@ set -euo pipefail
 UMBREL_ROOT="$(readlink -f $(dirname "${BASH_SOURCE[0]}")/../../..)"
 MOUNT_POINT="/mnt/data"
 EXTERNAL_UMBREL_ROOT="${MOUNT_POINT}/umbrel"
+SWAP_DIR="/swap"
+SWAP_FILE="${SWAP_DIR}/swapfile"
 
 check_root () {
   if [[ $UID != 0 ]]; then
@@ -169,11 +171,9 @@ main () {
   echo "Bind mounting external storage over local Umbrel installation..."
   mount --bind "${EXTERNAL_UMBREL_ROOT}" "${UMBREL_ROOT}"
 
-  echo "Bind mounting external storage to /swap"
-  systemctl stop dphys-swapfile
-  mkdir -p "${MOUNT_POINT}/swap" "/swap"
-  mount --bind "${MOUNT_POINT}/swap" "/swap"
-  systemctl start dphys-swapfile
+  echo "Bind mounting external storage to ${SWAP_DIR}"
+  mkdir -p "${MOUNT_POINT}/swap" "${SWAP_DIR}"
+  mount --bind "${MOUNT_POINT}/swap" "${SWAP_DIR}"
 
   echo "Bind mounting SD card root at /sd-card..."
   [[ ! -d "/sd-root" ]] && mkdir -p "/sd-root"
@@ -184,8 +184,15 @@ main () {
   sleep 1
   df -h "${UMBREL_ROOT}" | grep --quiet '/dev/sd'
 
-  echo "Checking /swap is now on external storage..."
-  df -h "/swap" | grep --quiet '/dev/sd'
+  echo "Checking ${SWAP_DIR} is now on external storage..."
+  df -h "${SWAP_DIR}" | grep --quiet '/dev/sd'
+
+  echo "Setting up swapfile"
+  rm "${SWAP_FILE}"
+  fallocate -l 2G "${SWAP_FILE}"
+  chmod 600 "${SWAP_FILE}"
+  mkswap "${SWAP_FILE}"
+  swapon "${SWAP_FILE}"
 
   echo "Checking SD Card root is bind mounted at /sd-root..."
   df -h "/sd-root${UMBREL_ROOT}" | grep --quiet "/dev/root"
--- a/scripts/umbrel-os/external-storage/mount
+++ b/scripts/umbrel-os/external-storage/mount
@@ -169,24 +169,24 @@ main () {
   echo "Bind mounting external storage over local Umbrel installation..."
   mount --bind "${EXTERNAL_UMBREL_ROOT}" "${UMBREL_ROOT}"
 
-  echo "Bind mounting SD card root at /sd-card..."
-  [[ ! -d "/sd-root" ]] && mkdir -p "/sd-root"
-  mount --bind "/" "/sd-root"
-
   echo "Bind mounting external storage to /swap"
   mkdir -p "${MOUNT_POINT}/swap" "/swap"
   mount --bind "${MOUNT_POINT}/swap" "/swap"
+
+  echo "Bind mounting SD card root at /sd-card..."
+  [[ ! -d "/sd-root" ]] && mkdir -p "/sd-root"
+  mount --bind "/" "/sd-root"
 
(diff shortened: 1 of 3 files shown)


lukechilds:
I've tested this and it's working for me.

You can test the entire update process with:

```
sudo mkdir /umbrel
sudo chown umbrel:umbrel /umbrel
git clone --branch swap https://github.com/lukechilds/umbrel.git /umbrel
sudo ~/umbrel/scripts/update/update --path /umbrel
```

After the next reboot you should have a 4GB swap file at `/swap/swapfile` which is bind mounted to external storage and the `dphys-swapfile` dep should be removed.

Do you think we should create the new swapfile on the fly after OTA update so users don't need to reboot to get access to swap?

scripts/umbrel-os/external-storage/mount
```diff
@@ -178,6 +184,16 @@ main () {
   sleep 1
   df -h "${UMBREL_ROOT}" | grep --quiet '/dev/sd'
 
+  echo "Checking ${SWAP_DIR} is now on external storage..."
+  df -h "${SWAP_DIR}" | grep --quiet '/dev/sd'
+
+  echo "Setting up swapfile"
+  rm "${SWAP_FILE}"
```
mayankchhabra: Looks like this can throw an error if there's no existing swapfile.

scripts/update/02-run.sh
```diff
@@ -12,4 +12,13 @@ echo "========= Stage: Post-update =========="
 echo "======================================="
 echo
 
-# Nothing here for now
+# Make Umbrel OS specific post-update changes
+if [[ ! -z "${UMBREL_OS:-}" ]]; then
+
+  # Remove dphys-swapfile since we now use our own swapfile logic
+  if command -v dphys-swapfile >/dev/null 2>&1; then
```
mayankchhabra, line 19: Do you think we can move this logic to `01-run.sh` instead? Right after this block of code:

https://github.com/getumbrel/umbrel/blob/8c5bf7aabd71450b1ebba6c959bac85a7e13a34d/scripts/update/01-run.sh#L32

mayankchhabra (commented):
Looks great, @lukechilds! Super-sleek. Will test it out now.

scripts/umbrel-os/external-storage/mount
mayankchhabra: Oh sweet, you fixed it already!

mayankchhabra:
The 4GB swap was created successfully with no problems in my testing! Although it looks like it didn't remove `dphys-swapfile` for some reason (hence the total swap size is 4.1G instead of 4.0G because of the existing 100mb swapfile):

```
umbrel@umbrel-testnet:~ $ free -h
              total        used        free      shared  buff/cache   available
Mem:          7.6Gi       881Mi       5.9Gi       8.0Mi       932Mi       6.7Gi
Swap:         4.1Gi          0B       4.1Gi
umbrel@umbrel-testnet:~ $ dphys-swapfile
Usage: /sbin/dphys-swapfile {setup|swapon|swapoff|uninstall}
```

lukechilds:
>Although it looks like it didn't remove dphys-swapfile for some reason

Do you still have the `dphys-swapfile` binary available?

mayankchhabra:
Yep, it looks like so: 

```
umbrel@umbrel-testnet:~ $ dphys-swapfile
Usage: /sbin/dphys-swapfile {setup|swapon|swapoff|uninstall}
```

mayankchhabra:
Lol also didn't realize your branch was at v0.2.2. The update process changed my installation from v0.2.3 to v0.2.2. A good stress test regardless haha!

scripts/update/02-run.sh
```diff
@@ -12,4 +12,13 @@ echo "========= Stage: Post-update =========="
 echo "======================================="
 echo
 
-# Nothing here for now
+# Make Umbrel OS specific post-update changes
+if [[ ! -z "${UMBREL_OS:-}" ]]; then
```
mayankchhabra, line 16: Ah okay, this could be the reason why `dphys-swapfile` wasn't removed locally on my Pi. I think we need to source `/etc/default/umbrel` first for `$UMBREL_OS` to be available.

lukechilds: Yeah it is, I just tested, I just ninja added that line in before you tested.

Although it seems a bit pointless to source `/etc/default/umbrel` and then test for the existence of a variable from the file we just sourced, when we could just test for the file.

I think there should be somewhere we can source `/etc/default/umbrel` that makes it available everywhere.

scripts/update/02-run.sh
lukechilds: It can definitely do there, I just thought it belonged in post-update.

Why do you think removing an old dependency should be part of the install stage instead of the post-update stage?

And what kind of logic do you imagine should be in the post-update stage?

lukechilds:
>Lol also didn't realize your branch was at v0.2.2. The update process changed my installation from v0.2.3 to v0.2.2. A good stress test regardless haha!

Oops, like like I forget to fetch upstream before I branched off.

Yeah `update --path` does no version checks, it just blindly executes the update scripts. This is useful for dev so you can repeatedly re-install your branch over itself while testing.

scripts/update/02-run.sh
lukechilds: Sourcing the file for now in this PR but I think we should create `/etc/profile.d/umbrel-os.sh` with:

```shell
export UMBREL_OS=$version
```

mayankchhabra: > Although it seems a bit pointless to source /etc/default/umbrel and then test for the existence of a variable from the file we just sourced, when we could just test for the file.

I agree, although the use-case right now is just testing for the variable, it can be testing for multiple variables as well as their values. For example, we can test if `$UMBREL_OS` is a specific version and then do something.

> Sourcing the file for now in this PR but I think we should create /etc/profile.d/umbrel-os.sh

That's a good idea. Are you sure the environment variables will be available for root though? Because we already source `/etc/default/umbrel` in `/etc/profile` but they're still not exposed to the root.

scripts/update/02-run.sh
mayankchhabra: Yeah tbh, it's objective I guess. I think this specific change of uninstalling a dependency can go alongside the manual creation and enabling of the swap file, which might be better suited at the install stage. Also, it may help to keep all Umbrel OS modifications in the same script/block so they're easier to just remove when we release an update that pins another release.

Eventually, we can end up dedicating the entire `01-run.sh` for Umbrel OS installation changes, and let it skip directly to `02-run.sh` for non-Umbrel OS systems. 

As for the post-update stage, that could be better suited for unit checks to make sure everything has been installed successfully.

mayankchhabra: Also, it could be a good idea to import your semver bash library or just wrap the swap logic inside:

if [[ $UMBREL_OS == "v0.1.2" ]] || [[ $UMBREL_OS == "v0.2.3" ]] {
  remove dphys-swapfile
  create a 4GB swap and enable it
}

scripts/update/02-run.sh
lukechilds: That could be cause we're not exporting them you need to do `export ENV=value` to ensure the shell session exports the variable to any child processes.

But no I'm not sure about root, I need to test it.

scripts/update/02-run.sh
lukechilds: I'm not sure we should limit with semver.

Checking that we are on Umbrel OS and if `dphys-swapfile` exists more accurately conveys what we're trying to do.

If for some reason they still had it installed and skipped an update, or we were to not accurately describe the version range, or mess up a future update version requirement, or they are on a commit version, or we're testing locally on a dev branch, are all scenarios where that could fail unpredictably.

Just saying "if we're on Umbrel OS and if this thing exists, uninstall it" is simpler, less error prone and more accurately describes what we are doing.

lukechilds:
@mayankchhabra I've made the suggested amendments.

If you re-install based on [this](https://github.com/getumbrel/umbrel/pull/180#issuecomment-679956638) and make sure you remove swapfile/bindmount and re-install adn enable dphys-swapfile you should see everything is uninstalled/re-created after update.

mayankchhabra (approved):
Tested and everything looks great, @lukechilds! Awesome work here. Let's release this tomorrow 🚀

```
umbrel@umbrel-testnet:~ $ dphys-swapfile
-bash: dphys-swapfile: command not found
umbrel@umbrel-testnet:~ $ free -h
              total        used        free      shared  buff/cache   available
Mem:          7.6Gi       384Mi       6.9Gi       8.0Mi       376Mi       7.1Gi
Swap:         4.0Gi          0B       4.0Gi
```

PR #186, opened 2020-08-27
Sync filesystem changes before checking partition exists

lukechilds:
It seems in some strange edge cases the device file can exist (`/dev/sda`) but not the partition file (`/dev/sda1`).

If we don't run `sync` before checking for the existence of the Umbrel partition, it can appear to be non-existent and we will erroneously wipe the external storage and reformat it, despite it containing a valid Umbrel install.

Running `sync` before the check ensures the latest filesystem changes are showing properly.


mayankchhabra (approved):
Great find, what a pain that'd have been to debug and test haha.

PR #194, opened 2020-08-30
Remove stale statuses/lock files

mayankchhabra:
This helps us avoid a situation where a stale lock/status file prevents the execution of a trigger forever. For now, we're removing `backup-in-progress`.


lukechilds (approved):
I think this is fine as a quick hack for now but just making a note that long term we should improve the reliability of the lock file creation with something like discussed here: https://github.com/getumbrel/umbrel/pull/188#discussion_r479740992

PR #196, opened 2020-08-30
Fix incorrect time calculation

lukechilds:
Sooooo it turns out there's actually 60 minutes in an hour, not 10.

### DIFF
--- a/scripts/backup/decoy-trigger
+++ b/scripts/backup/decoy-trigger
@@ -28,7 +28,7 @@ check_dependencies shuf
 main () {
   while true; do
     minutes_in_seconds="60"
-    hours_in_seconds="$((${minutes_in_seconds} * 10))"
+    hours_in_seconds="$((60 * ${minutes_in_seconds}))"
     max_interval="$((8 * ${hours_in_seconds}))"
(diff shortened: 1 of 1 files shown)


meeDamian:
> there's 60 minutes in an hour

Source?

meeDamian (approved):

nolim1t:
> > there's 60 minutes in an hour
> 
> Source?

https://en.wikipedia.org/wiki/Hour

mayankchhabra (approved):
giphy (2)

junderw:
ACKSHUALLY BY SOME DEFINITIONS THERE ARE EXACTLY 59.999999837260128 SECONDS IN A MINUTE.

meeDamian:
Someone kick @lukechilds out, he's clearly either malicious, incompetent, or a North Corean spy 😱

sindresorhus:
According to the theory of relatively, the existing calculation is correct if you observe an hour on earth from a place with different gravitational field, also called gravitational time dilation, which makes me suspect that @lukechilds is currently not residing on earth.

lukechilds:
>which makes me suspect that @lukechilds is currently not residing on earth.

Bruh, don't ruin my opsec.

PR #211, opened 2020-08-31
Persist LND backend between OTA updates

lukechilds:
Replaces #208 

This removes the possibility of pre-existing install being re-configured back to Neutrino and then never switching back to Bitcoin Core.

### DIFF
--- a/scripts/configure
+++ b/scripts/configure
@@ -25,7 +25,7 @@ fi
 check_dependencies docker docker-compose dirname readlink
 
 # Switch to Umbrel's root directory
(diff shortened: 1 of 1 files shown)


mayankchhabra (approved):
Tested and all good. Great catch and fix!

nolim1t:
Wonder if we can just query `getblockchaininfo` to bitcoind and getting the sync information and handling it instead? Would this be a better approach to use instead?

lukechilds:
That would be quite error prone since it could fail if Bitcoin Core is not running (if it's been taken down for an OTA update), end then the node would be rolled back to Neutrino mode and stuck there forever after the first OTA update.

PR #213, opened 2020-09-02
Make backups thread safe

lukechilds:
This change removes the need for a lock file to prevent concurrent backups. Multiple backup threads can now run concurrently without issue.

This improves reliability, since int he edge case that a lock file doesn't get cleaned up correctly, no future backups would ever run.

### DIFF
--- a/scripts/backup/backup
+++ b/scripts/backup/backup
@@ -3,9 +3,9 @@
 set -euo pipefail
 
 UMBREL_ROOT="$(readlink -f $(dirname "${BASH_SOURCE[0]}")/../..)"
-BACKUP_FOLDER=".backup/$RANDOM"
-BACKUP_ROOT="${UMBREL_ROOT}/${BACKUP_FOLDER}"
-BACKUP_FILE="${BACKUP_FOLDER}/backup.tar.gz.pgp"
+BACKUP_FOLDER_NAME="backup"
+BACKUP_ROOT="${UMBREL_ROOT}/.backup/$RANDOM/${BACKUP_FOLDER_NAME}"
+BACKUP_FILE="${BACKUP_ROOT}/backup.tar.gz.pgp"
 
 check_dependencies () {
   for cmd in "$@"; do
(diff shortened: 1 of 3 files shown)


scripts/backup/backup
```diff
@@ -3,9 +3,10 @@
 set -euo pipefail
 
 UMBREL_ROOT="$(readlink -f $(dirname "${BASH_SOURCE[0]}")/../..)"
-BACKUP_FOLDER="backup"
-BACKUP_ROOT="${UMBREL_ROOT}/${BACKUP_FOLDER}"
-BACKUP_FILE="${UMBREL_ROOT}/backup.tar.gz.pgp"
+BACKUP_ROOT="${UMBREL_ROOT}/.backup/$RANDOM"
```
mayankchhabra, line 6: Wow, didn't know there was an always available `$RANDOM` in shell 😍

mayankchhabra (approved):
Looks great, @lukechilds! Quick question — should we remove `${UMBREL_ROOT}/statuses/backup-in-progress` via the next OTA update if it exists (just as a cleanup so everything's consistent with repo, even though it wouldn't make a difference)?

scripts/backup/backup
lukechilds: It's not very random, in fact it's not even uniformly random, but yeah, it's always available in a shell and fine for stuff like this.

It's also not actually a variable but a function, so you can use it multiple times in the same line

e.g:

```
$ echo "${RANDOM}:${RANDOM}:${RANDOM}"
16448:13088:7550
```

lukechilds:
>should we remove ${UMBREL_ROOT}/statuses/backup-in-progress via the next OTA update if it exists (just as a cleanup so everything's consistent with repo, even though it wouldn't make a difference)?

Good idea! I'll add a line to the update script.

mayankchhabra (approved):

lukechilds:
So it turns out the start script would've cleared the backup lock file anyway.

So I've removed all instances of cleaning up the lock file that were left in since we don't need them anymore, and added a single cleanup in the OTA update script.

PR #214, opened 2020-09-02
Reduce decoy backup frequency

lukechilds:
I had originally incorrectly assumed the LND channel.backup changes with each channel state change so we needed lots of decoys to reduce channel state change information leakage.

However the file only changes on channel open/close so we can safely reduce the decoy backup frequency.


mayankchhabra (approved):
Looks good!

PR #215, opened 2020-09-02
Handle upload fail

lukechilds:
Currently if the upload fails the backup script exits uncleanly.

This PR correctly handles a failed upload by allowing the script to clean up and prints the error in the logs.

### DIFF
--- a/scripts/backup/backup
+++ b/scripts/backup/backup
@@ -116,7 +116,11 @@ if [[ $BITCOIN_NETWORK == "regtest" ]]; then
 fi
 
 echo "Uploading backup..."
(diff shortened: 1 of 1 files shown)


mayankchhabra:
Looks good! Makes me wonder if we should add a while loop and retry uploading the backup for like 5 or 10 times before exiting the script?

lukechilds:
If we're going to retry we should probably do that in a a wrapper around the entire backup logic to avoid a race condition.

It opens the possibility of getting in a retry loop due to a network failure, a separate backup gets triggered, networking now works again, the new backup uploads, then our retry works and we overwrite the latest backup with an outdated backup.

So we should rebuild the backup file from the latest files on disk.

Maybe we can merge this as is and look into that as a separate PR.

mayankchhabra (approved):
Yeah, that makes sense!

PR #225, opened 2020-09-07
Increase nginx proxy timeout to support more intensive computations

nolim1t:
Increase NGINX proxy timeout to support cool shit like https://github.com/getumbrel/umbrel-middleware/pull/60

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -17,7 +17,7 @@ http {
 
 	access_log  /etc/nginx/log/access.log  main;
 	sendfile        on;
-	keepalive_timeout  65;
-
-	 include /etc/nginx/conf.d/*.conf;	
+	keepalive_timeout  360;
+    proxy_read_timeout 600;
+    include /etc/nginx/conf.d/*.conf;	
 }


nginx/nginx.conf
```diff
@@ -17,7 +17,7 @@ http {
 
 	access_log  /etc/nginx/log/access.log  main;
 	sendfile        on;
-	keepalive_timeout  65;
-
-	 include /etc/nginx/conf.d/*.conf;	
+	keepalive_timeout  360;
+    proxy_read_timeout 600;
+    include /etc/nginx/conf.d/*.conf;
```
jgmontoya, line 22: Is there any rationale for those numbers or where they decided just based on trying and getting `gettxoutsetinfo` to run?

Not relevant but consistent indentation :man_shrugging: :
```suggestion
	keepalive_timeout  360;
	proxy_read_timeout 600;
	include /etc/nginx/conf.d/*.conf;	
```

mayankchhabra:
Thanks @nolim1t. I don't really think it's a good idea as increased timeouts can have their own side-effects. Also see https://github.com/getumbrel/umbrel-middleware/pull/60#issuecomment-691425143.

nginx/nginx.conf
nolim1t: Also might fix other gateway timeouts

nolim1t:
Might fix other gateway timeout issues too

mayankchhabra:
Closed by https://github.com/getumbrel/umbrel-middleware/pull/60

PR #235, opened 2020-09-15
Update license

mayankchhabra:


wiz:
Did all the contributors to the Umbrel open source project consent to re-license their contributions under this new license before you merged the PR? I'm not a lawyer but I don't think you can unilaterally change the licensing of an open source project like this. Here are some examples of how some other projects did it:
https://github.com/rust-bitcoin/rust-lightning/issues/659
https://github.com/mempool/mempool/pull/388

AaronDewes:
Contributors were @dennisreimann @nolim1t @mayankchhabra @lukechilds and @vindard until then.
I agree with the change, @nolim1t was part of Umbrel, not sure if he was at the time of the change. @mayankchhabra @lukechilds are still developing Umbrel, I agree with the change. @dennisreimann did only a minor documentation change, so if he doesn't agree with it, we can revert that. @vindard did a more important contrbution, so you're right, @vindard hasn't agrreed with that change. @vindard what is your opinion? And also, @dennisreimann.

AaronDewes:
Also, AFAIK the MIT license does allow changing the license or distributing a fork under a more restrictive license, so it shouldn't be an issue.

nolim1t:
> Contributors were @dennisreimann @nolim1t @mayankchhabra @lukechilds and @vindard until then.
> 
> I agree with the change, @nolim1t was part of Umbrel, not sure if he was at the time of the change. @mayankchhabra @lukechilds are still developing Umbrel, I agree with the change. @dennisreimann did only a minor documentation change, so if he doesn't agree with it, we can revert that. @vindard did a more important contrbution, so you're right, @vindard hasn't agrreed with that change. @vindard what is your opinion? And also, @dennisreimann.

I'm all for more permissive licensing.

When I created this project it was very permissive.

No idea why it was changed when clearly it wasn't broken.

dennisreimann:
I agree with @nolim1t and just recently learned, that the Umbrel codebase isn't completely MIT-licensed anymore. 

No hard feelings, but as a developer and also user I prefer an open and permissionless approach and think [@raucao is spot on here](https://github.com/getumbrel/umbrel/issues/291#issuecomment-800190433).

Issue #205, opened 2020-08-31
OTA update stuck at "downloading"


mayankchhabra:
This has, in all cases, been reported by users on manual Umbrel installations.

jgmontoya:
I'm having this issue as well. Also manual installation. ¿How should one go about updating without losing information?

snapo:
same here, manual installation (ubuntu 20.04) upgrade from 26 or so to 28 stuck at downloading ... as i have no funds yet on it i play a bit around. maybe i figure it out

mayankchhabra:
As a temporary fix, if your update is stuck at downloading, you can run `cd` into the umbrel directory and run the following to reset:

```
rm statuses/update-status.json && echo "{\"state\": \"success\", \"progress\": 100, \"description\": \"\", \"updateTo\": \"\"}" >> statuses/update-status.json
```

jgmontoya:
If anyone want's to update on the meantime this worked for me: `sudo ./scripts/update/update --repo getumbrel/umbrel#v0.2.8`

I don't think it matters but at least I had my instance stopped at the time of updating.

snapo:
> If anyone want's to update on the meantime this worked for me:
> `sudo ./scripts/update/update --repo getumbrel/umbrel#v0.2.8`
> 
> I don't think it matters but at least I had my instance stopped at the time of updating.

That worked absolutely superb...

jgmontoya:
I could update using the dashboard normally from v0.2.8 to v0.2.9 without any problem whatsoever.

I think this issue can be closed :tada:

mayankchhabra:
Well, that makes it more complex because I can replicate the issue on http://testnet.getumbrel.com lol which runs a manual Umbrel installation on Ubuntu 18.04.

An interesting discovery is that the OTA update doesn't log to `syslog` when it's triggered through the UI gets stuck. And if you try to redirect the log output to another file, the update doesn't get stuck anymore. Crazy.

mayankchhabra:
Still the same issue on https://testnet.getumbrel.com when trying to update to v0.2.10:

![image](https://user-images.githubusercontent.com/10330103/92639112-fe457f00-f2f8-11ea-8011-3e05d3fa9049.png)

lukechilds:
>If anyone want's to update on the meantime this worked for me:
>`sudo ./scripts/update/update --repo getumbrel/umbrel#v0.2.8`
>
>I don't think it matters but at least I had my instance stopped at the time of updating.

[...]

PR #239, opened 2020-09-19
Check for invalid 'RELEASE' values in update script

vindard:
This adds two checks for invalid `RELEASE` values that can potentially come from the `update-status.json` file:

- check that `RELEASE` is not an empty value. This can happen if for example the update script was run when 
the `update-status.json` file is in its default success state and the `"updateTo"` value is an empty string

- check that `RELEASE` points to a valid URL to avoid any further error from attempting to do operations on an invalid curl call

 ---

### Error details
These changes would help the `update` script to not change the `update-status.json` to an invalid `Downloading...` state if [`Ln106`](https://github.com/getumbrel/umbrel/blob/836026624b843114d180e9860b79b7f217ed7915/scripts/update/update#L106) runs and the script fails, which then blocks the dashboard from loading permanently (see issue #205).

#### on unexpected `update` script run

- the `update` script errors as follows

    image

- the update status file changes from 'success' to 'installing'

    image

 - the dashboard becomes inaccessible and shows this Downloading message incorrectly instead

    image

### DIFF
--- a/scripts/update/update
+++ b/scripts/update/update
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -euo pipefail
+
 if [[ $UID != 0 ]]; then
     echo "Update script must be run as root"
     exit 1
--- a/scripts/update/update
+++ b/scripts/update/update
@@ -17,12 +17,7 @@ check_dependencies () {
 }
 
 curl_check () {
-  status=$(curl -L --head --silent ${1} | grep 'HTTP.*200')
-  if echo $status | grep -q 200; then
-    return 0
-  else
-    return 1
-  fi
+  curl -LIs -o /dev/null -w "%{http_code}" "$1" | grep -q 200
 }
 
 check_dependencies jq curl rsync
--- a/scripts/update/update
+++ b/scripts/update/update
@@ -14,6 +14,15 @@ check_dependencies () {
   done
 }
 
+curl_check () {
+  status=$(curl -L --head --silent ${1} | grep 'HTTP.*200')
+  if echo $status | grep -q 200; then
+    return 0
+  else
+    return 1
+  fi
+}
+
 check_dependencies jq curl rsync
 
 UMBREL_ROOT="$(readlink -f $(dirname "${BASH_SOURCE[0]}")/../..)"
@@ -59,6 +68,11 @@ elif [[ "${update_type}" == "path" ]]; then
   RELEASE=$(cat "$update_path"/info.json | jq .version -r)
 fi
 
+if [[ -z "${RELEASE}" ]]; then
+  echo "Empty release version found in update-status.json file, aborting..."
+  exit 1
+fi
+
 echo
 echo "======================================="
 echo "=============== UPDATE ================"
@@ -71,6 +85,14 @@ echo
 if [[ -f "$UMBREL_ROOT/statuses/update-in-progress" ]]; then
     echo "An update is already in progress. Exiting now."
     exit 2
+fi
+
+RELEASE_URL="https://github.com/getumbrel/umbrel/archive/$RELEASE.tar.gz"
+
+# Check that release url is valid
+if ! curl_check "${RELEASE_URL}"; then
+  echo "Could not reach any release files hosted at ${RELEASE_URL}"
+  exit 1
 fi
 
 echo "Creating lock"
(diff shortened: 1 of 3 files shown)


scripts/update/update
```diff
@@ -14,6 +14,15 @@ check_dependencies () {
   done
 }
 
+curl_check () {
+  status=$(curl -L --head --silent ${1} | grep HTTP/1.1 | tail -n 1)
```
AaronDewes, line 18: I'm not sure why you're using HTTP/1.1 here, I think just HTTP would also work, in the very unlikely event that GitHub would switch to requiring HTTP2 without announcing it before, this check would always fail, resulting in users being unable to install updates.

vindard: Good catch. I was thinking that too, but the tricky thing was I wasn't sure if the string `HTTP` could also appear anywhere in the body of the response that would mess up the `200` check in the lines below. Maybe exclusively grep-ing for an expression like `HTTP.*200` as a success indicator could help with this :thinking:

vindard:
**Update**

In response to [this comment](https://github.com/getumbrel/umbrel/issues/205#issuecomment-695361223) on the issue, I still haven't been able to replicate the in-flight update failure.

Given this though, I tried a `try-catch` of sorts on the entire update script with a `catch` script to place the `update-status.json` file back to a consistent state and cleanup up any dangling update dirs/files if the `update` script errors in any way. All changes are entirely contained in [this commit](https://github.com/getumbrel/umbrel/pull/239/commits/cbe186ad0d762981bf3586828356f16da2e8ec29) and can be backed out or changed if the approach isn't ideal

mayankchhabra:
@vindard this is awesome! It also saves our face in case we accidentally merge an updated `info.json` with a new release tag, but forget to release the tarball on GitHub lol.

I really like your [catch script](https://github.com/vindard/umbrel/blob/42b8385243f8000ae5fa36c0e0d05c3b3d6eaeee/scripts/update/catch). We've been thinking of implementing `trap` for better error handling like you pointed out, but haven't had a chance to look at it.

> On testing I found that the update-status.json did not play well with the front-end transitions if the catch is triggered

If you change `"state": "errored"` to `"state": "failed"` [here](https://github.com/vindard/umbrel/blob/42b8385243f8000ae5fa36c0e0d05c3b3d6eaeee/scripts/update/catch#L36) it would work 🎉

In my testing it worked flawlessly. The only change I had to make was to move the curl check inside the following `if` block:

https://github.com/getumbrel/umbrel/blob/f17c563ed948c34c98daf47a74f6b3b6eb0db4fe/scripts/update/update#L111

scripts/update/update
meeDamian:
```sh
curl_check () {
	curl -LIs -o /dev/null -w "%{http_code}" "$1" | grep -q 200
}
```

Should work too

vindard: Oh wow this is wizardry, thanks for the suggestion man! :raised_hands: ... just committed here as well

vindard:
> In my testing it worked flawlessly. The only change I had to make was to move the curl check inside the following `if` block:

Oh ya true, only want that running when `update_type` is `"ota"` :see_no_evil: 

Awesome, glad those changes worked out. I went ahead and:
- made the ~`errored`~`failed` change
- pulled across the `catch` script back into this branch, and
- made that `curl_check` move change

This branch should be good to go for testing again now I think :v: 

---

Also with the `trap`, when I checked it out some more it looked like it's used more for being more versatile with the types of errors/exits you can catch. I'm guessing the generic `||` version I implemented here might be ok as a generic catch for what we're trying to do with the `update` script?

_Let me know if not though and I can check out the `trap` some more and implement that instead_

scripts/update/update
```diff
@@ -87,10 +100,17 @@ EOF
 
 # Download new release
 if [[ "${update_type}" == "ota" ]]; then
+  # Check that release url is valid
+  if ! curl_check "${RELEASE_URL}"; then
```
lukechilds, line 104: Is this downloading the update and piping it to `/dev/null` to check the response code, then downloading it again later to extract?

lukechilds: Couldn't we just handle this in one go with:
```bash
if ! curl -L "${RELEASE_URL}" | tar -xz --strip-components=1; then
  # handle error
fi
```

vindard: > Is this downloading the update and piping it to /dev/null to check the response code, then downloading it again later to extract?

`curl_check` just gets the headers (`/dev/null` is to throw away all the header boilerplate and `-w "%{http_code}"` is to get just the last response code)

> Couldn't we just handle this in one go

Oh yea this totally works as well (tested to double-check) and is much cleaner. Ya sure I'll make the change to this instead :ok_hand:

mayankchhabra:
Thanks a lot for working on this @vindard! As you might have noticed, quite a lot has changed recently with Umbrel 0.5 😅. Closing this PR now, and my apologies for the resulting unproductive hard work. :(

vindard:
> Thanks a lot for working on this @vindard! As you might have noticed, quite a lot has changed recently with Umbrel 0.5 😅. Closing this PR now, and my apologies for the resulting unproductive hard work. :(

Lol! Ya that would've been one hell of a merge conflict to resolve 😂

PR #240, opened 2020-09-20
Add a '--all' flag to stop script for background processes

vindard:
### Description

This PR adds a `--all` flag to the `scripts/stop` script which looks for the background scripts that currently get kicked off in Umbrel and kills them with a `-9` sigterm.

This option could be useful when doing dev and debug work on a manual install and shouldn't affect the other scripts that use the `stop` script internally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,12 @@ To stop Umbrel, run:
 sudo ./scripts/stop
 ```
 
+To stop background processes as well, use the `--all` flag:
+
+```bash
+sudo ./scripts/stop --all
+```
+
 ## 🎹 Services orchestrated
 
 - [`Umbrel Dashboard`](https://github.com/getumbrel/umbrel-dashboard)
--- a/scripts/start
+++ b/scripts/start
@@ -92,8 +92,27 @@ if [[ ! -z "${hidden_service_url:-}" ]]; then
     echo "  http://${hidden_service_url}"
 fi
 
+
 # Store Process Group ID for killing all child processes later
-PGID_FILE="${UMBREL_ROOT}/start_pgid.log"
 PGID=$(ps -o pgid= $$ | grep -o '[0-9]*')
-echo $PGID > ${PGID_FILE}
-echo "Start script process group ID: ${PGID}"
+PGID_FILE="${UMBREL_ROOT}/start_pgid.log"
+
+pgid_processes () {
+  ps -e -o pgid,pid | awk -v p=$1 '$1 == p {print $2}'
+}
+
+save_new_pgid () {
+  echo $PGID > ${PGID_FILE}
+  echo "Start script process group ID: ${PGID}"
+}
+
+if [[ -e "${PGID_FILE}" ]]; then
+  EXISTING_PGID=$(cat ${PGID_FILE} | head -n 1)
+  if [[ -z "$(pgid_processes ${EXISTING_PGID})" ]]; then
+    save_new_pgid
+  else
+    echo "Existing PGID with running processes found"
+  fi
+else
+  save_new_pgid
+fi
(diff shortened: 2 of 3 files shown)


mayankchhabra:
Thanks for the PR, @vindard! This would indeed be very useful for development purposes. 

In my testing, it seems to stop the background processes as intended, except one process (`fswatch`) which is spawned by `karen`:

```
root      1996     1  0 10:19 pts/0    00:00:00 fswatch -0 --event=PlatformSpecific /home/umbrel/vindard/events/signals
```

Also, if we're going this route, do you think it's a better idea to store `pid` for background processes in the start script (maybe in a new directory `$UMBREL_ROOT/pids/`) and then kill processes based off those `pids`?

nolim1t:
> Thanks for the PR, @vindard! This would indeed be very useful for development purposes.
> 
> In my testing, it seems to stop the background processes as intended, except one process (`fswatch`) which is spawned by `karen`:
> 
> ```
> root      1996     1  0 10:19 pts/0    00:00:00 fswatch -0 --event=PlatformSpecific /home/umbrel/vindard/events/signals
> ```
> 
> Also, if we're going this route, do you think it's a better idea to store `pid` for background processes in the start script (maybe in a new directory `$UMBREL_ROOT/pids/`) and then kill processes based off those `pids`?

I think storing all pids can be useful for a convenient way of checking for stuff running

vindard:
> Also, if we're going this route, do you think it's a better idea to store pid for background processes in the start script (maybe in a new directory $UMBREL_ROOT/pids/) and then kill processes based off those pids?

Stopping by tracking PIDs instead of trying a string match makes a lot of sense 

I did some research and found a great way to do this by using the "process group ID" of the start script to be able to isolate and kill all child processes spawned ([see here](https://stackoverflow.com/questions/392022/whats-the-best-way-to-send-a-signal-to-all-members-of-a-process-group/15139734#15139734)). With this method, we'd only need to store the `PGID` from the `start` script and then use that to kill all processes in the `stop` script. On my testing so far this seems to work well.

I'm currently persisting the PGID in a text file in the `scripts` folder which gets deleted when the `stop` script is run with the `--all` flag. The only current shortcoming is if someone runs the `start` script again when it has already been run then the legit PGID value which spawned currently running child processes gets overwritten.

lukechilds:
I haven't tested this out yet but just a couple of observations:

Instead of the default behaviour leaving processes running and having an `--all` flag to kill them, it probably makes more sense for the default behaviour to kill everything and have a `--preserve-update` flag that prevents killing the update script.

When a user is stopping Umbrel, they will always want to stop everything. The only time we ever don't want to stop everything is during an OTA update when we want to stop Umbrel without killing the update script. And even then, we want to stop as much as possible, ideally also `karen`, just not the `update` job it just spawned. Then after OTA the user is running up to date background services.

>I did some research and found a great way to do this by using the "process group ID" of the start script to be able to isolate and kill all child processes spawned (see here). With this method, we'd only need to store the PGID from the start script and then use that to kill all processes in the stop script. On my testing so far this seems to work well.

If the background services are already running, they won't be spawned again. So if you save the PGID of the latest `start` run, but that didn't spawn any background jobs because they were already running, I don't think killing based on the PGID would work.

lukechilds:
In fact, come to think of it, maybe we don't need a flag at all.

We can just always kill all background processes.

We probably never want to kill `karen` triggers mid execution, so we can just implement something in `karen` that disowns its child processes.

stevenroose:
@vindard just a heads up that if you don't have explicit individual approval by all the contributors to this repository to publish your modified source code ("Adapted Material"), you are officially in violation of the license of this project and you can be sued:

mayankchhabra:
Thanks again @vindard! 🙏

PR #242, opened 2020-09-22
Integrate electrs

lukechilds:
This integrates the electrs Electrum server into Umbrel.

Related:
- https://github.com/getumbrel/umbrel-manager/pull/49
- https://github.com/getumbrel/umbrel-dashboard/pull/221

You can connect Electrum Wallet to Umbrel with:

```
your-electrum-hidden-service.onion:50001:t
```

You can verify you're connected by viewing the console tab in Electrum and you should see the text "Umbrel v0.2.10".

Screenshot 2020-09-19 at 22 08 01

I don't seem to be able to get ARM builds of the electrs Docker images working in GitHub Actions via QEMU, so for now I've manually built the Docker images locally and pushed them to Docker Hub via CLI.


docker-compose.yml
```diff
@@ -168,6 +168,21 @@ services:
...
+        electrs:
+              container_name: electrs
+              image: getumbrel/electrs:v0.8.5
+              logging: *default-logging
+              depends_on: [ bitcoin ]
+              volumes:
+                - ${PWD}/bitcoin:/data/.bitcoin:ro
+                - ${PWD}/electrs:/data
+              restart: unless-stopped
+              stop_grace_period: 1m
```
mayankchhabra, line 180: I've not tried it out, but do you think `1m` is enough or shall we increase this to `5m` or so?

mayankchhabra (commented):
Looks great, @lukechilds! Excited to try this out! 

Also, we probably need to update `.updateinclude` (to include the electrum config file) and `.updateignore` (to ignore everything else).

docker-compose.yml
lukechilds: I've not noticed it being force killed but yeah no harm in increasing it a little to be safe.

mayankchhabra (approved):

PR #254, opened 2020-10-01
Update documentation - Raspberry PI 3 should no longer be supported, amend to RPI4 or 64-bit linux system

nolim1t:
Because Electrum is a new requirement and because of the RAM requirements. remove references to `raspberry pi 3`

Also add disk space requirements too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,8 @@ We run it on Raspbery Pis (ARMv8) as a part of [Umbrel OS](https://github.com/ge
 
 If you're looking to run Umbrel on:
 
-- A Raspberry Pi 3 or 4 (recommended) - [Download Umbrel OS](https://github.com/getumbrel/umbrel-os)
+- A Raspberry Pi 4 (recommended) - [Download Umbrel OS](https://github.com/getumbrel/umbrel-os)
+- 1 TB of disk space (for bitcoin blockchain and electrum files)
 - Anything else (**not recommended** as it's experimental at the moment) - [Install Umbrel](#-installation)
 
 ## 🛠 Installation
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ We run it on Raspbery Pis (ARMv8) as a part of [Umbrel OS](https://github.com/ge
 
 If you're looking to run Umbrel on:
 
-- A Raspberry Pi 4 (recommended) - [Download Umbrel OS](https://github.com/getumbrel/umbrel-os)
+- A Raspberry Pi 4 with at least 2 Gb RAM - [Download Umbrel OS](https://github.com/getumbrel/umbrel-os) or a 64-bit Linux based system with at least 2 Gb of RAM
 - 1 TB of disk space (for bitcoin blockchain and electrum files)
 - Anything else (**not recommended** as it's experimental at the moment) - [Install Umbrel](#-installation)
 


nolim1t:
Opinion @mayankchhabra

README.md
```diff
@@ -21,7 +21,8 @@ We run it on Raspbery Pis (ARMv8) as a part of [Umbrel OS](https://github.com/ge
 
 If you're looking to run Umbrel on:
 
-- A Raspberry Pi 3 or 4 (recommended) - [Download Umbrel OS](https://github.com/getumbrel/umbrel-os)
+- A Raspberry Pi 4 (recommended) - [Download Umbrel OS](https://github.com/getumbrel/umbrel-os)
```
AaronDewes, line 24: I think you should also add something like this, some people might have an older 1GB version.

```suggestion
- A Raspberry Pi 4 with at least 2GB of RAM (recommended) - [Download Umbrel OS](https://github.com/getumbrel/umbrel-os)
```

nolim1t: Cool thanks for that

lukechilds: I think maybe also drop `(recommended)` since we are no longer recommending one of two options.

nolim1t: Thats actually a good point, I've amended that

README.md
```diff
@@ -21,7 +21,8 @@ We run it on Raspbery Pis (ARMv8) as a part of [Umbrel OS](https://github.com/ge
 
 If you're looking to run Umbrel on:
 
-- A Raspberry Pi 3 or 4 (recommended) - [Download Umbrel OS](https://github.com/getumbrel/umbrel-os)
+- A Raspberry Pi 4 with at least 2 Gb RAM - [Download Umbrel OS](https://github.com/getumbrel/umbrel-os) or a 64-bit Linux based system with at least 2 Gb of RAM
```
mayankchhabra, line 24: Can you please update this to:

```suggestion
Raspberry Pi 4 with at least 2GB RAM (recommended) - [Download Umbrel OS](https://github.com/getumbrel/umbrel-os)
```

Umbrel OS is not compatible with any "64-bit Linux based system with at least 2 Gb of RAM" so we can't add that. Also, the "recommended" is required since we compare it with "Anything else" (down below).

README.md
```diff
@@ -21,7 +21,8 @@ We run it on Raspbery Pis (ARMv8) as a part of [Umbrel OS](https://github.com/ge
...
+- 1 TB of disk space (for bitcoin blockchain and electrum files)
```
mayankchhabra, line 25: This isn't required here as all the requirements/instructions are listed on the umbrel-os.
```suggestion

```

mayankchhabra (changes requested):

README.md
mayankchhabra: We show "recommended" in comparison to "anything else":

image

README.md
nolim1t: What about for installs which don't use umbrel-os

README.md
nolim1t: True that Umbrel OS isn't compatible but this is compatible with anything and there are installs where people are just taking this.

README.md
lukechilds: Ahh, yes, missed that.

lukechilds:
Replaced via https://github.com/getumbrel/umbrel/pull/259

PR #255, opened 2020-10-01
Expose Bitcoin Core RPC

lukechilds:
Resolves #247 

This PR exposes Bitcoin Core RPC via a Tor hidden service.

It also modifies the update and configure scripts to ensure they preserve the RPC credentials between OTA updates and configuration script runs.

Related:
- https://github.com/getumbrel/umbrel-manager/pull/51
- https://github.com/getumbrel/umbrel-dashboard/pull/231

PR #256, opened 2020-10-01
Allow the user to specify a NGINX port

### LINKED ISSUE #253: Allow NGINX to run on a custom port
I have my own RP running on port 80/443 already that handles all SSL termination.

When umbrel starts up it tries to bind to port 80 which is already taken.

Maybe some option to stop Nginx binding to the host's port 80 would be helpful. 

For now, I could work around it by commenting out these lines in docker-compose:
```
                # ports:
                #     - "80:80"
```

nolim1t:
This issue addresses and fixes #253


docker-compose.yml
```diff
@@ -30,7 +30,7 @@ services:
                 restart: unless-stopped
                 stop_grace_period: 30s
                 ports:
-                    - "80:80"
+                    - "${WEBPORT:-80}:80"
```
lukechilds, line 33:
```suggestion
                    - "${UMBREL_HTTP_PORT:-80}:80"
```

lukechilds (changes requested):
Thanks @nolim1t this looks great!

Just added a suggested change for the env var to keep it inline with the style of the existing ones.

docker-compose.yml
nolim1t: Ok

mayankchhabra:
Maybe we should add the new variable to ".env" for it to persist between OTA updates.

lukechilds:
Yeah good point.

@nolim1t You can see examples in the configure script of how to persist settings in `.env`.

nolim1t:
> Maybe we should add the new variable to ".env" for it to persist between OTA updates.

I allowed for both. Default to 80 if not set.

lukechilds:
If users modify the `.env` file directly, it'll be wiped after an OTA update.

See an example of how this needs to be done in the configure script.

Or see this PR for an example of how we added `BITCOIN_RPC_AUTH `: https://github.com/getumbrel/umbrel/pull/255/files

nolim1t:
Hmm what's the point of the .env file if you are wiping it each time.

May as well just keep it in the compose file 😛 

Better to just add what's needed and use default values if nothing is set.

nolim1t:
I've added `UMBREL_HTTP_PORT` to the configure script too.. it will [persist to `.env' if this set](https://github.com/getumbrel/umbrel/pull/256/commits/df7dbcea55de191add284504b47f1aa66ceb3484). Otherwise it will just stick with the defaults as per docker-compose.yml config.

nolim1t:
Also added how to use this setting to the README file

scripts/configure
```diff
@@ -198,6 +198,11 @@ if [[ -f "${STATUS_DIR}/node-status-bitcoind-ready" ]]; then
   sed -i "s/bitcoin.node=.*/bitcoin.node=bitcoind/g;" "$LND_CONF_FILE"
 fi
 
+# If there was an UMBREL_HTTP_PORT set before then add it to the $ENV_FILE
+if [ ! -z $UMBREL_HTTP_PORT ]; then
```
lukechilds, line 202: This will throw an unbound variable error if `UMBREL_HTTP_PORT` is not set.

lukechilds (changes requested):

scripts/configure
nolim1t: ok ill remove it

mayankchhabra (changes requested):
Thanks, @nolim1t. 

It'd be great to stick to using the same pattern that we use to set environment variables in `.env` instead of echoing it directly, i.e. by having `UMBREL_HTTP_PORT=<port>` in [`templates/.env-sample`](https://github.com/getumbrel/umbrel/blob/master/templates/.env-sample) and setting it via a `sed` in the config script.

nolim1t:
> Thanks, @nolim1t.
> 
> It'd be better if we can stick to the same pattern that we use to set environment variables in `.env` instead of echoing it directly, i.e. by having `UMBREL_HTTP_PORT=<port>` in [`templates/.env-sample`](https://github.com/getumbrel/umbrel/blob/master/templates/.env-sample) and setting it via a `sed` in the config script.

does the following changes work?

louneskmt (commented):
You've set NGINX_PORT in the `.env` file but UMBREL_HTTP_PORT in all the other files.

nolim1t:
Thanks well spotted!

README.md
```diff
@@ -57,6 +57,9 @@ sudo NETWORK=testnet ./scripts/start
 
 # For regtest, run:
 sudo NETWORK=regtest ./scripts/start
+
+# For a custom HTTP_PORT (example 8080)
+sudo UMBREL_HTTP_PORT=8080 ./scripts/start
```
louneskmt, line 62:
```suggestion
sudo NGINX_PORT=8080 ./scripts/start
```

docker-compose.yml
```diff
@@ -30,7 +30,7 @@ services:
                 restart: on-failure
                 stop_grace_period: 30s
                 ports:
-                    - "80:80"
+                    - "${UMBREL_HTTP_PORT:-80}:80"
```
louneskmt, line 33:
```suggestion
                    - "${NGINX_PORT:-80}:80"
```

louneskmt (changes requested):
There were two left 😁

nolim1t:
Thanks again lol

maltokyo:
Awesome to see this coming! Finally, support for being behind a reverse proxy.

nolim1t:
> Awesome to see this coming! Finally, support for being behind a reverse proxy.

yep just need someone to review it and we are good

junderw:
I've added NGINX_PORT to my .env file in anticipation for this update :-D

lukechilds:
Closed by https://github.com/getumbrel/umbrel/pull/562

PR #257, opened 2020-10-01
Add 200MB offset to RAM check

lukechilds:
We add a 200MB offset because the 2GB Pi actually only has ~1.9GB of RAM and was failing this check.


mayankchhabra:
Looks good! 

Did you accidentally commit the backup signal file? 😅

lukechilds:
Woops yep, we should probably add the signals dir to gitignore.

mayankchhabra (approved):

Issue #206, opened 2020-08-31
Log to umbrel_root/logs


mayankchhabra:
Create a new `/logs` directory in `$UMBREL_ROOT` and log to it instead of `syslog`. It helps prevent unnecessary SD card wear and tear on a Raspberry Pi and also enables us to easily request relevant log files from users.

nmfretz:
Closing as completed.

PR #262, opened 2020-10-03
Redirect OTA update's logs to a designated log file

mayankchhabra:
This seems to fix https://github.com/getumbrel/umbrel/issues/205 (please don't ask me how or why cuz I have no clue lol).

Also, since we're making changes to the parent update script, it wouldn't be executed on the next immediate OTA update, but on all the updates post it. It means that users will still encounter issue #205 one more time on the next update (the last time, I hope).

The new `logs/` directory can further be used to store logs from other background processes and docker containers going forward (#206).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ db/*
 electrs/*
 events/signals/*
 lnd/*
+logs/*
 statuses/*
 tor/*
 
@@ -33,5 +34,6 @@ tor/*
 !db/.gitkeep
 !events/signals/.gitkeep
 !lnd/.gitkeep
+!logs/.gitkeep
 !tor/data/.gitkeep
 !tor/run/.gitkeep
--- a/scripts/start
+++ b/scripts/start
@@ -68,6 +68,10 @@ export COMPOSE_HTTP_TIMEOUT=240
 
 cd "$UMBREL_ROOT"
 
+echo "Rotating log files..."
+echo
+rm -rf logs/*
+
 echo "Starting karen..."
 echo
 ./karen &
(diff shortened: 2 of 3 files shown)


lukechilds:
>This seems to fix #205 (please don't ask me how or why cuz I have no clue lol).

Lol. Pretty cryptic but what's happening is when people manually run the `start` script and then close the shell session (or exit SSH) the file descriptors (stdout/stderr) no longer exist. So any scripts that get fired as a sub process of the start script (basically any event triggered by karen) that tries to write to stdout or stderr fails.

This is why OTA update's are failing on most custom Umbrel installs. Users are probably running the start script and logging out of SSH. As soon as the OTA update script gets triggered and needs stdout/stderr it fails. On Umbrel OS we launch the start script via systemd and [pipe stdout/stderr to syslog](https://github.com/getumbrel/umbrel/blob/771affe0e595ff1b46f58898562de8f758fbd1b0/scripts/umbrel-os/services/umbrel-startup.service#L21-L22) so the file descriptors always exist.

If users with custom installs launch the start script with `sudo ./start &> output.log`, (pipe stdout and stderr to a file called output.log) everything will work with the current setup.

While the solution in this PR does allow OTA updates to work without issue, we should probably handle this better directly in the start script. It could also cause other scripts to fail. I'd suggest explicitly piping the output of all child processes of the start script to their own log file. I've submitted a PR here: https://github.com/getumbrel/umbrel/pull/269

lukechilds (changes requested):
See comment

mayankchhabra:
Damn, that makes so much sense in retrospect. Great job debugging this, @lukechilds! 🙌

PR #263, opened 2020-10-03
Prevent unresolved update status on a failed OTA update

mayankchhabra:
This PR helps prevents situations like #205 in the future, where the cause of the error is unknown and the update-status.json is forever stuck in an unresolved state (i.e. anything except `success` or `failed`).


lukechilds (approved):
Great idea. Tested and working well.

PR #264, opened 2020-10-03
Add a link to Discord

AaronDewes:
Adds a link to the unofficial Umbrel Discord server.


mayankchhabra:
Thanks, @AaronDewes! 

I do agree that having an unofficial community is a great idea as it has its own benefits, especially around censorship-resistant communication that's free from our control, but I don't think we should actively promote it through the official repos for two reasons:

1. We shouldn't be the mediators that vouch for unofficial communities as it defeats the benefits of non-affiliation.
2. The more chat options we provide, the more we confuse and segment the user-base.

PR #269, opened 2020-10-05
Fix missing file descriptors

### LINKED ISSUE #205: OTA update stuck at "downloading"

lukechilds:
Resolves #205 
Replaces #262 

See https://github.com/getumbrel/umbrel/pull/262#issuecomment-703765275 for reasoning:

>>This seems to fix #205 (please don't ask me how or why cuz I have no clue lol).
>
>Lol. Pretty cryptic but what's happening is when people manually run the `start` script and then close the shell session (or exit SSH) the file descriptors (stdout/stderr) no longer exist. So any scripts that get fired as a sub process of the start script (basically any event triggered by karen) that tries to write to stdout or stderr fails.
>
>This is why OTA update's are failing on most custom Umbrel installs. Users are probably running the start script and logging out of SSH. As soon as the OTA update script gets triggered and needs stdout/stderr it fails. On Umbrel OS we launch the start script via systemd and [pipe stdout/stderr to syslog](https://github.com/getumbrel/umbrel/blob/771affe0e595ff1b46f58898562de8f758fbd1b0/scripts/umbrel-os/services/umbrel-startup.service#L21-L22) so the file descriptors always exist.
>
>If users with custom installs launch the start script with `sudo ./start &> output.log`, (pipe stdout and stderr to a file called output.log) everything will work with the current setup.
>
>While the solution in this PR does allow OTA updates to work without issue, we should probably handle this better directly in the start script. It could also cause other scripts to fail. I'd suggest explicitly piping the output of all child processes of the start script to their own log file. I've submitted a PR here: https://github.com/getumbrel/umbrel/pull/269

### DIFF
--- a/scripts/start
+++ b/scripts/start
@@ -72,15 +72,15 @@ cd "$UMBREL_ROOT"
 
 echo "Starting karen..."
 echo
-./karen &> "${UMBREL_LOGS}/karen" &
+./karen &>> "${UMBREL_LOGS}/karen" &
 
 echo "Starting backup monitor..."
 echo
-./scripts/backup/monitor &> "${UMBREL_LOGS}/backup-monitor" &
+./scripts/backup/monitor &>> "${UMBREL_LOGS}/backup-monitor" &
 
 echo "Starting decoy backup trigger..."
 echo
-./scripts/backup/decoy-trigger &> "${UMBREL_LOGS}/backup-decoy-trigger" &
+./scripts/backup/decoy-trigger &>> "${UMBREL_LOGS}/backup-decoy-trigger" &
 
 echo
 echo "Starting Docker services..."
--- a/scripts/start
+++ b/scripts/start
@@ -29,6 +29,8 @@ check_dependencies fswatch
 check_dependencies rsync jq curl
 
 UMBREL_ROOT="$(dirname $(readlink -f "${BASH_SOURCE[0]}"))/.."
+UMBREL_LOGS="${UMBREL_ROOT}/logs"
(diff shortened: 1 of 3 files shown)


scripts/start
```diff
@@ -29,6 +29,8 @@ check_dependencies fswatch
 check_dependencies rsync jq curl
 
 UMBREL_ROOT="$(dirname $(readlink -f "${BASH_SOURCE[0]}"))/.."
+UMBREL_LOGS="${UMBREL_ROOT}/logs"
+mkdir -p "${UMBREL_LOGS}"
```
mayankchhabra, line 33: Do you think this is required since we're explicitly committing the logs directory?

mayankchhabra (commented):
Looks great! Awesome work, @lukechilds! 👏

scripts/start
lukechilds: I think it's good to have, ideally we should always ensure folders exist before writing to them.

Otherwise if someone does `rm -rf logs/` the Umbrel service will fail to start.

mayankchhabra: I agree, but isn't that the same for every other directory? Like if someone did `rm -rf electrs/` or `rm -rf bitcoin/`, it would fail to start too (and/or result in errors)?

lukechilds: Yeah I suppose. I added this before checking `logs/` into git so I guess it's fine to remove it.

mayankchhabra (approved):
Tested and worked flawlessly! Finally, we'll have OTA updates working again for everyone 🎉

PR #273, opened 2020-10-06
Fix #272

### LINKED ISSUE #272: Fix GitHub Actions OS image jobs
We have a GitHub Action to build an Umbrel OS image based on a PR branch or master on each push.

However, the job builds Umbrel OS based on the tip of that branch, not the current commit.

Steps to reproduce:

- Add a commit to a branch
- Push
- The Umbrel OS job will start building for that commit
- Add another commit
- Push
- The first job is still building the OS
- By the time the job gets round to pulling in the Umbrel repo it pulls the tip of the branch
- This job is for the first commit, but it's building Umbrel OS based on the second commit (branch tip)

If the first commit was ok, but the second commit broke builds, the job would fail on the first commit which is misleading.

To resolve this we should pass the commit hash into the Umbrel OS build and make sure the job for that commit builds Umbrel OS based on exactly that getumbrel/umbrel commit and not the tip of that branch.

AaronDewes:


AaronDewes:
This requires https://github.com/getumbrel/umbrel-os/pull/134 to work properly.

.github/workflows/pull_request.yml
```diff
@@ -37,12 +37,14 @@ jobs:
...
         IMAGE_NAME="umbrel-os-dev"
+        REPO="$(curl -s https://api.github.com/repos/getumbrel/umbrel/pulls/${PR} | jq .head.repo.full_name | sed 's/"//g')"
         UMBREL_REPO="$(curl -s https://api.github.com/repos/getumbrel/umbrel/pulls/${PR} | jq .head.repo.clone_url | sed 's/"//g')"
-        UMBREL_BRANCH="$(curl -s https://api.github.com/repos/getumbrel/umbrel/pulls/${PR} | jq .head.ref | sed 's/"//g')"
-        echo "::set-env name=UMBREL_OS_VERSION::$UMBREL_OS_VERSION"
-        echo "::set-env name=IMAGE_NAME::$IMAGE_NAME"
-        echo "::set-env name=UMBREL_REPO::$UMBREL_REPO"
-        echo "::set-env name=UMBREL_BRANCH::$UMBREL_BRANCH"
+        BRANCH="$(curl -s https://api.github.com/repos/getumbrel/umbrel/pulls/${PR} | jq .head.ref | sed 's/"//g')"
+        UMBREL_BRANCH="$(curl https://api.github.com/repos/$REPO/branches/$BRANCH | jq .commit.sha | sed 's/"//g')"
```
lukechilds, line 43: This is vulnerable to the same bug as #272, but with a reduced time interval.

If you push two commits soon after each other, the first action could start after the second commit, meaning when you query the GitHub API for the tip of the branch, you get the same commit for both builds.

Instead of querying the API you can just use the preset GitHub Actions environment variables.

https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables

`GITHUB_SHA` should be the commit SHA that triggered the workflow which is exactly what we want.

lukechilds (changes requested):

AaronDewes:
Completely broke everything, including file permissions... Will fix it soon

AaronDewes:
GITHUB_SHA on PRs is a merge commit, this merge commit ensures that even if the PR branch misses the latest commits, they are in the workflow, but the latest version should work around that.

AaronDewes:
This version now detects the commit correctly.

.github/workflows/pull_request.yml
```diff
@@ -8,6 +8,25 @@ jobs:
...
+
+    - name: Setting env vars
+      env:
+        PR: ${{ github.event.number }}
+      run: |
+        UMBREL_OS_VERSION="dev"
+        IMAGE_NAME="umbrel-os-dev"
+        UMBREL_REPO="$(curl -s https://api.github.com/repos/getumbrel/umbrel/pulls/${PR} | jq .head.repo.clone_url | sed 's/"//g')"
+        TMP=$(git show HEAD | head -n 2 | tail -n 1)
+        UMBREL_BRANCH=${TMP:15:7}
```
lukechilds, line 22: What are the 15:7 values here for?

.github/workflows/pull_request.yml
```diff
@@ -8,6 +8,25 @@ jobs:
...
+    - uses: actions/checkout@v1
+
+    - name: Setting env vars
+      env:
+        PR: ${{ github.event.number }}
+      run: |
+        UMBREL_OS_VERSION="dev"
+        IMAGE_NAME="umbrel-os-dev"
+        UMBREL_REPO="$(curl -s https://api.github.com/repos/getumbrel/umbrel/pulls/${PR} | jq .head.repo.clone_url | sed 's/"//g')"
+        TMP=$(git show HEAD | head -n 2 | tail -n 1)
```
lukechilds, line 21: Is this to get the branch? Is that a reliable query?

Running that in a local repo I get:

```
Author: Luke Childs <lukechilds123@gmail.com>
```

If you're getting current branch name here I'd suggest:

```
git branch --show-current
```

lukechilds (changes requested):

AaronDewes:
The way GitHub action PR builds work ist this: 
1. GitHub creates a merge commit that doesn't exist I the repo, but is only used to make sure the PR is tested against the latest version of the repo.
2. This merge commit is cloned (I'm not sure from where)

The code I wrote basically gets the merge commit, takes the line that shows which commits are merged (You get the commit author if the commit is not a mere commit, but in this case, It always is one), then it removes the text "Merge" and the second commit that I the current commit in Umbrel and so outputs the real commit to clone, the 15 and the 7 are used for that. I don't remember the exact details right now, but I think it takes 7 letters from the text after the 15th letter

lukechilds:
Ahh ok, thanks for the explanation.

Is there are build or something where you logged the output to see how to grab the value? If so could you link it to me so I can take a look?

lukechilds:
Alternatively would getting the commit hash before the merge commit work? `git rev-parse HEAD^1`

AaronDewes:
> Alternatively would getting the commit hash directly preceding the merge commit work?
> 
> ```
> git rev-parse HEAD^1
> ```

No, the needed commit is at least two commits behind the merge commit, but depending on how much the PR branch missed, it might be more, so this method doesn't work.

AaronDewes:
This is the TMP var in a test build: https://github.com/AaronDewes/umbrel/runs/1353844339?check_suite_focus=true#step:3:17

That it shows
that @mayankchhabra's repo is being used is because it only takes the PR number and then looks for the author of the PR with that number in the official Umbrel repo, where I this case @mayankchhabra authored the PR.

AaronDewes:
The old way to set env vars will be removed November 16th, please fix at least these commands am or merge this PR until then, otherwise the builds will fail

lukechilds (approved):

lukechilds:
Thank you @AaronDewes! Apologies for the delay in reviewing.

PR #288, opened 2020-10-15
Make sure containers don't autostart on boot

lukechilds:
We only want containers to restart if they error. The current setting means Docker will autostart the containers on each boot which we never want to do.

This can actually cause pretty nasty bugs if the OS tries to launch the containers and mount volumes that are supposed to exist on external storage devices before the storage device has yet been mounted.


mayankchhabra (approved):
Looks great!

PR #293, opened 2020-10-19
Update apt packages on update

AaronDewes:
Even if most of Umbrel runs in Docker containers, the main system should have all security updates installed.

### DIFF
--- a/scripts/update/01-run.sh
+++ b/scripts/update/01-run.sh
@@ -27,16 +27,6 @@ if [[ ! -z "${UMBREL_OS:-}" ]]; then
     echo "============================================="
     echo
 
-    # In Umbrel OS v0.1.2, we need to bind Avahi to only
-    # eth0,wlan0 interfaces to prevent hostname cycling
-    # https://github.com/getumbrel/umbrel-os/issues/76
-    # This patch can be safely removed from Umbrel v0.3.x+
-    if [[ $UMBREL_OS == "v0.1.2" ]] && [[ -f "/etc/avahi/avahi-daemon.conf" ]]; then
-        echo "Binding Avahi to eth0 and wlan0"
-        sed -i "s/#allow-interfaces=eth0/allow-interfaces=eth0,wlan0/g;" "/etc/avahi/avahi-daemon.conf"
-        systemctl restart avahi-daemon.service
-    fi
-
     # Update SD card installation
     if  [[ -f "${SD_CARD_UMBREL_ROOT}/.umbrel" ]]; then
         echo "Replacing ${SD_CARD_UMBREL_ROOT} on SD card with the new release"
@@ -56,8 +46,10 @@ if [[ ! -z "${UMBREL_OS:-}" ]]; then
     fi
     
     # Update apt packages on update
-    # Tis checks if the package will be installed from buster-security and if yes, installs it.
-    apt-get -s dist-upgrade | grep "^Inst" | grep -i securi | awk -F " " {'print $2'} | xargs apt-get install
+    DEBIAN_FRONTEND=noninteractive apt-get install unattended-updates -y
+    # Manual run of the update (Normally for debugging purposes only, but we don't want to have a potential backdoor in Umbrel)
+    # https://wiki.debian.org/UnattendedUpgrades#Manual_run_.28for_debugging.29
+    unattended-upgrade -d
 fi
 
 cat <<EOF > "$UMBREL_ROOT"/statuses/update-status.json
--- a/scripts/update/01-run.sh
+++ b/scripts/update/01-run.sh
@@ -28,14 +28,14 @@ if [[ ! -z "${UMBREL_OS:-}" ]]; then
     echo
 
     # In Umbrel OS v0.1.2, we need to bind Avahi to only
-    # eth0,wlan0 interfaces to prevent hostname cycling	
-    # https://github.com/getumbrel/umbrel-os/issues/76	
-    # This patch can be safely removed from Umbrel v0.3.x+	
-    if [[ $UMBREL_OS == "v0.1.2" ]] && [[ -f "/etc/avahi/avahi-daemon.conf" ]]; then	
-        echo "Binding Avahi to eth0 and wlan0"	
-        sed -i "s/#allow-interfaces=eth0/allow-interfaces=eth0,wlan0/g;" "/etc/avahi/avahi-daemon.conf"	
-        systemctl restart avahi-daemon.service	
-    fi	
+    # eth0,wlan0 interfaces to prevent hostname cycling
+    # https://github.com/getumbrel/umbrel-os/issues/76
+    # This patch can be safely removed from Umbrel v0.3.x+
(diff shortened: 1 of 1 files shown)


mayankchhabra (commented):
Looks good to me. Thoughts, @lukechilds?

lukechilds:
Great idea @AaronDewes!

We should definitely do security critical OS updates during OTA, however a couple of issues with this implementation:
- Always prefer apt-get over apt in scripts
- `upgrade` is not safe for a system upgrade, you want `dist-upgrade` to make sure dependencies are correctly resolved
- We probably don't want to blindly upgrade everything, it would be better to just stick to security critical upgrades for more predictability.

I'd recommend using `unattended-upgrade`. It's a better tool for this job: https://wiki.debian.org/UnattendedUpgrades

We might need to manually whitelist some important packages from the Pi sources if they aren't whitelisted by default.

lukechilds (changes requested):
Great idea to do this manually as part of the OTA process, it shouldn't be automated.

However see comment for some issues with implementation: https://github.com/getumbrel/umbrel/pull/293#issuecomment-721801306

lukechilds:
It's also worth noting that this is a potential attack vector. It essentially gives Debian package publishers a backdoor into Umbrel.

However it's significantly more likely that an Umbrel is compromised due to running outdated software with security vulnerabilities, than a Debian package maintainer managing to slip a vulnerability past all other contributors that directly targets Umbrel. So for that reason I think it's a good tradeoff.

We obviously already trust Debian package maintainers to some extent since we haven't manually verified all Debian packages when building the initial Umbrel OS image.

AaronDewes:
> We probably don't want to blindly upgrade everything, it would be better to just stick to security critical upgrades for more predictability.

AFAIK Debian doesn't really publish big updates that could break anything, but mostly fixes, and bigger updates only to web browsers, so I don't think much things will break, because Umbrel OS always uses Buster, not stable, but yes, I think `unattended-upgrade` is better, I'll try to implement it on the weekend

AaronDewes:
This new version should work.

scripts/update/01-run.sh
```diff
@@ -54,6 +54,14 @@ if [[ ! -z "${UMBREL_OS:-}" ]]; then
         echo "ERROR: No Umbrel installation found at SD root ${SD_CARD_UMBREL_ROOT}"
         echo "Skipping updating on SD Card..."
     fi
+    
+    # Configure unattended-upgrades
+    # This can be safely removed from Umbrel v0.3.x+
+    echo "Configuring automated updates of apt packages"
+    apt-get update
+    DEBIAN_FRONTEND=noninteractive apt-get install unattended-upgrades -y
+    echo unattended-upgrades unattended-upgrades/enable_auto_updates boolean true | debconf-set-selections
```
lukechilds, line 63: So this will run in the background automatically, right?

I don't think we should do that since it's essentially a back door. This is the reason we don't automatically update Umbrel already and require the user to confirm first.

We should do security updates only during OTA so users who don't want to opt in to trusting third parties don't have software automatically installed on their hardware.

We should run `unattended-upgrade` once during OTA and ensure it isn't running automatically in the background.

scripts/update/01-run.sh
```diff
@@ -54,6 +54,14 @@ if [[ ! -z "${UMBREL_OS:-}" ]]; then
         echo "ERROR: No Umbrel installation found at SD root ${SD_CARD_UMBREL_ROOT}"
         echo "Skipping updating on SD Card..."
     fi
+    
+    # Configure unattended-upgrades
+    # This can be safely removed from Umbrel v0.3.x+
+    echo "Configuring automated updates of apt packages"
+    apt-get update
+    DEBIAN_FRONTEND=noninteractive apt-get install unattended-upgrades -y
```
lukechilds, line 62: It would be nice to wrap this in a check to see if the package is already installed to suppress extra unneeded log output.

We should also add this dep to Umbrel OS.

lukechilds (changes requested):
Sorry for the delay reviewing this Aaron!

I'd love to see this merged but a couple of tiny adjustments are required.

I'm focused on getting open PRs merged at the moment so I wont leave you hanging for ages if you resolve this.

Thanks!

scripts/update/01-run.sh
AaronDewes: But why can't we just use apt then? I think unattended-updates was made to be used automatically.

lukechilds: Because `unattended-upgrade` only updates security critical dependencies.

AaronDewes: `apt-get -s dist-upgrade | grep "^Inst" | grep -i securi | awk -F " " {'print $2'} | xargs apt-get install`

That should work too, and isn't an additional dependency

lukechilds: Wouldn't this have the potential to match packages with the string "securi" as part of their name?

I think it would be more reliable to use `unattended-upgrade` for this which is purpose built for this use case.

AaronDewes: Okay, I'll do that

scripts/update/01-run.sh
```diff
@@ -27,16 +27,6 @@ if [[ ! -z "${UMBREL_OS:-}" ]]; then
     echo "============================================="
     echo
 
-    # In Umbrel OS v0.1.2, we need to bind Avahi to only
```
lukechilds, line 30: Are you able to undo this deletion?

We normally keep one PR down to a single atomic change and squash it down into a single commit when we merge to keep the commit history clean.

scripts/update/01-run.sh
```diff
@@ -54,6 +44,12 @@ if [[ ! -z "${UMBREL_OS:-}" ]]; then
         echo "ERROR: No Umbrel installation found at SD root ${SD_CARD_UMBREL_ROOT}"
         echo "Skipping updating on SD Card..."
     fi
+    
+    # Update apt packages on update
+    DEBIAN_FRONTEND=noninteractive apt-get install unattended-updates -y
```
lukechilds, line 49: Have you checked this definitely won't setup automatic background updates by default? There's no extra step we need to take to disable that?

Also would be great to guard this off with a check that only runs the install if the dep isn't already installed.

scripts/update/01-run.sh
```diff
@@ -54,6 +44,12 @@ if [[ ! -z "${UMBREL_OS:-}" ]]; then
...
+    # Manual run of the update (Normally for debugging purposes only, but we don't want to have a potential backdoor in Umbrel)
+    # https://wiki.debian.org/UnattendedUpgrades#Manual_run_.28for_debugging.29
+    unattended-upgrade -d
```
lukechilds, line 52: It would be nice to update the status file here saying something like "Running security updates" so the user gets some feedback on this step, it could take a while.

lukechilds (changes requested):
Looking good Aaron, just some tiny tweaks.

scripts/update/01-run.sh
AaronDewes: From the documentation it seems like it only runs when configured to do so, I'll add a check

lukechilds (approved):
Great stufff!

Thanks Aaron.

PR #294, opened 2020-10-22
Add debug script

AaronDewes:
If users send the output of `umbrel-debug` with a bug report, it should help to find the problem by providing the information that is often needed. (CPU temperature, Power issues, Network config, swap usage, RPi model, firmware version, Docker logs and the version of Umbrel and Umbrel OS)

### DIFF
--- a/scripts/debug
+++ b/scripts/debug
@@ -1,8 +1,7 @@
 #!/bin/bash -e
 
-# Based on raspinfo: https://github.com/raspberrypi/utils/blob/master/raspinfo/raspinfo
-
-# Copyright: 2018-2019, Raspberry Pi (Trading) Ltd.
+# This script is based on raspinfo: https://github.com/raspberrypi/utils/blob/master/raspinfo/raspinfo
+# raspibfo is Copyright: 2018-2019, Raspberry Pi (Trading) Ltd.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
@@ -27,6 +26,7 @@
 # LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING 
 # NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS 
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+# This fork is Copyright: 2020 Umbrel.
 
 [[ -f "/etc/default/umbrel" ]] && source "/etc/default/umbrel"
 
--- a/scripts/debug
+++ b/scripts/debug
@@ -98,11 +98,11 @@ cat /proc/swaps
 
 
 echo
-echo "Networking information"
-echo "----------------------"
+echo "Karen logs"
+echo "----------"
 echo
 
-ifconfig | sed -e "s/\([0-9a-fA-F]\{1,4\}:\)\{7,7\}[0-9a-fA-F]\{1,4\}/y.y.y.y.y.y.y.y/g" | sed -e "s/[0-9a-fA-F]\{1,4\}:\(:[0-9a-fA-F]\{1,4\}\)\{1,4\}/y::y.y.y.y/g" | sed -e "s/\([0-9]\{1,3\}\.\)\{3,3\}[0-9]\{1,3\}/x.x.x.x/g" | sed -e "s/\([0-9a-fA-F]\{2,2\}\:\)\{5,5\}[0-9a-fA-F]\{2,2\}/m.m.m.m/g"
+tail -n 50 logs/karen.log | sed '/onion/d'
 
 echo
 echo "Docker containers"
@@ -157,6 +157,7 @@ if [[ ! -z "${UMBREL_OS:-}" ]]; then
         if [[ "${1}" == "--upload" ]]; then
             echo "You can also share this link instead:"
             ./scripts/debug | sed '/onion/d' | haste
+            dmesg | haste
         else
             echo "Run this script again with the --upload flag to automatically generate a link to share."
         fi
@@ -174,8 +175,9 @@ fi
 if [[ "${1}" == "--upload" ]]; then
     # This runs the script twice, but it works
     echo "This script could not automatically detect an issue with your Umbrel."
-    echo "Please share the following link and paste it in the Umbrel Telegram group (https://t.me/getumbrel) so we can help you with your problem."
+    echo "Please share the following links and paste it in the Umbrel Telegram group (https://t.me/getumbrel) so we can help you with your problem."
     ./scripts/debug | sed '/onion/d' | haste
+    dmesg | haste
 else
     echo "This script could not automatically detect an issue with your Umbrel."
     echo "Please copy the entire output of this script and paste it in the Umbrel Telegram group (https://t.me/getumbrel) so we can help you with your problem."
--- a/scripts/debug
+++ b/scripts/debug
@@ -35,6 +35,7 @@ echo "====================="
 
 UMBREL_ROOT="$(dirname $(readlink -f "${BASH_SOURCE[0]}"))/.."
 cd "${UMBREL_ROOT}"
+
 echo
 echo "Umbrel version"
 echo "---------------"
@@ -142,6 +143,8 @@ if [[ ! -z "${UMBREL_OS:-}" ]]; then
         echo "==== Result ===="
         echo "================"
         echo "You have either no external drive connected to your Raspberry Pi or the drive is not being detected."
+        echo "Please try to shutdown your Raspberry Pi ('sudo shutdown'), plug the SSD into the other USB3 port."
+        echo "After you've finished that, boot your Raspberry Pi back up."
         exit 0
     fi
     if [[ $no_of_block_devices -gt 1 ]]; then
@@ -158,7 +161,7 @@ echo "================"
 if [[ ! -z "${UMBREL_OS:-}" ]]; then
     if [[ $(vcgencmd get_throttled) != "throttled=0x0" ]]; then
         echo "There is probably an issue with either your power supply or the temperature of your Raspberry Pi is too high."
-        echo $(vcgencmd get_throttled)
+        vcgencmd get_throttled
         echo 'If the value above starts with "0x5", it is probably a power issue, otherwise it is probably a temperature issue.'
         exit 0
     fi
--- a/scripts/debug
+++ b/scripts/debug
@@ -28,6 +28,7 @@
 # NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS 
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+[[ -f "/etc/default/umbrel" ]] && source "/etc/default/umbrel"
 
 echo "====================="
 echo "= Umbrel debug info ="
@@ -166,5 +167,14 @@ if [[ ! -z "${UMBREL_OS:-}" ]]; then
         exit 0
     fi
 fi
-echo "This script could not automatically detect an issue with your Umbrel."
-echo "Please copy the entire output of this script and paste it in the Umbrel Telegram group (https://t.me/getumbrel) so we can help you with your problem."
+
+haste() { a=$(cat); curl -X POST -s -d "$a" https://hastebin.com/documents | awk -F '"' '{print "https://hastebin.com/"$4}'; }
+if [[ "${1}" == "--upload" ]]; then
+    # This runs the script twice, but it works
+    echo "This script could not automatically detect an issue with your Umbrel."
+    echo "Please share the following link and paste it in the Umbrel Telegram group (https://t.me/getumbrel) so we can help you with your problem."
+    ./scripts/debug | haste
+else
+    echo "This script could not automatically detect an issue with your Umbrel."
+    echo "Please copy the entire output of this script and paste it in the Umbrel Telegram group (https://t.me/getumbrel) so we can help you with your problem."
+fi
--- a/scripts/debug
+++ b/scripts/debug
@@ -70,6 +70,11 @@ if [[ ! -z "${UMBREL_OS:-}" ]]; then
     echo "----------"
 
     vcgencmd get_throttled
+
+    echo "Startup status"
+    echo "--------------"
+    systemctl status umbrel-startup
+    systemctl status umbrel-external-storage
 fi
 echo
 echo "Filesystem information"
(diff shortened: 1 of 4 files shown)


AaronDewes:
The script now also outputs a short help text that tells the user what they can do to fix their issue.

mayankchhabra (changes requested):
Thank you so much for this, @AaronDewes! Having a debug script would be extremely helpful. 

Can we please remove the need to install `smartmontools` since we aim to keep the host system as free from additional dependencies as possible?

Also, it'd be great to make this script available on both manual and Umbrel OS installations. So we can move it to `$umbrel/scripts/debug`, check for the existence of `$UMBREL_OS` variable and print additional info (related to throttling, external storage, etc).

A disk space check would be very useful. Couple of users who ran Umbrel on a 500GB drive ran out of space and had to debug their way out of it.

It'd also be helpful to fetch the status/logs of the `umbrel-startup` and `umbrel-external-storage` systemd services on Umbrel OS.

scripts/umbrel-os/debug
```diff
@@ -0,0 +1,142 @@
...
+echo
+
+ifconfig | sed -e "s/\([0-9a-fA-F]\{1,4\}:\)\{7,7\}[0-9a-fA-F]\{1,4\}/y.y.y.y.y.y.y.y/g" | sed -e "s/[0-9a-fA-F]\{1,4\}:\(:[0-9a-fA-F]\{1,4\}\)\{1,4\}/y::y.y.y.y/g" | sed -e "s/\([0-9]\{1,3\}\.\)\{3,3\}[0-9]\{1,3\}/x.x.x.x/g" | sed -e "s/\([0-9a-fA-F]\{2,2\}\:\)\{5,5\}[0-9a-fA-F]\{2,2\}/m.m.m.m/g"
+
+echo
+echo "Docker logs"
+echo "-----------"
+
+cd /home/umbrel/umbrel
+docker-compose logs
```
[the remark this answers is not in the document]
AaronDewes: Added that

AaronDewes:
I've fixed the things you mentioned now. The output of `df` already was in filesystem information, but I added -h to make it more readable. I haven't tested the script yet, could you please do that? (Not at home right now, did the edits from my phone)

scripts/debug
```diff
@@ -0,0 +1,144 @@
...
+# A PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE HOLDERS OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, 
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, 
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR 
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF 
+# LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING 
+# NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS 
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+[[ -f "/etc/default/umbrel" ]] && source "/etc/default/umbrel
```
louneskmt, line 31:
```suggestion
[[ -f "/etc/default/umbrel" ]] && source "/etc/default/umbrel"
```
Missing quotation mark here.

scripts/debug
```diff
@@ -0,0 +1,144 @@
...
+
+    vcgencmd measure_temp
+
+    echo
+    echo "Throttling"
+    echo "----------"
+
+    vcgencmd get_throttled
+
+    echo "Startup status"
```
louneskmt, line 74:
```suggestion

    echo
    echo "Startup status"
```
Just following the same style

AaronDewes:
Thanks, @louneskmt. Applied your suggestions.

scripts/debug
```diff
@@ -0,0 +1,146 @@
...
+echo "----------------------"
+echo
+
+ifconfig | sed -e "s/\([0-9a-fA-F]\{1,4\}:\)\{7,7\}[0-9a-fA-F]\{1,4\}/y.y.y.y.y.y.y.y/g" | sed -e "s/[0-9a-fA-F]\{1,4\}:\(:[0-9a-fA-F]\{1,4\}\)\{1,4\}/y::y.y.y.y/g" | sed -e "s/\([0-9]\{1,3\}\.\)\{3,3\}[0-9]\{1,3\}/x.x.x.x/g" | sed -e "s/\([0-9a-fA-F]\{2,2\}\:\)\{5,5\}[0-9a-fA-F]\{2,2\}/m.m.m.m/g"
+
+echo
+echo "Docker logs"
+echo "-----------"
+
+docker-compose logs --tail=100
```
mayankchhabra, line 100: On second thought, it might be a better idea to individually fetch logs of specific containers because one container could disproportionately spam the entire output of `docker-compose logs`. 

Plus we haven't really ever had the need to look at the logs of middleware, manager, nginx and dashboard containers so we can just rely on their statuses.

```suggestion
docker ps --format="table {{.Names}}\t{{.Status}}"

echo
echo "Bitcoin Core logs"
echo "-----------"
echo
docker-compose logs --tail=30 bitcoin

echo
echo "LND logs"
echo "-----------"
echo
docker-compose logs --tail=30 lnd

echo
echo "Tor logs"
echo "-----------"
echo
docker-compose logs --tail=30 tor
```

scripts/debug
```diff
@@ -0,0 +1,146 @@
...
+echo "================"
+if [[ ! -z "${UMBREL_OS:-}" ]]; then
+    if [[ $(vcgencmd get_throttled) != "throttled=0x0" ]]; then
+        echo "There is probably an issue with either your power supply or the temperature of your Raspberry Pi is too high."
+        echo $(vcgencmd get_throttled)
+        echo 'If the value above starts with "0x5", it is probably a power issue, otherwise it is probably a temperature issue.'
+        exit 0
+    fi
+fi
+echo "The script could not automatically detect an issue with your Raspberry Pi."
```
mayankchhabra, line 145:
```suggestion
echo "This script could not automatically detect an issue with your Umbrel."
```

scripts/debug
```diff
@@ -0,0 +1,146 @@
...
+echo "Please share the output of this command in the Umbrel Telegram group, and we will help you resolve the problem."
```
mayankchhabra, line 146:
```suggestion
echo "Please copy the entire output of this script and paste it in the Umbrel Telegram group (https://t.me/getumbrel) so we can help you with your problem."
```

scripts/debug
```diff
@@ -0,0 +1,146 @@
...
+
+echo
+echo "Networking Information"
+echo "----------------------"
+echo
+
+ifconfig | sed -e "s/\([0-9a-fA-F]\{1,4\}:\)\{7,7\}[0-9a-fA-F]\{1,4\}/y.y.y.y.y.y.y.y/g" | sed -e "s/[0-9a-fA-F]\{1,4\}:\(:[0-9a-fA-F]\{1,4\}\)\{1,4\}/y::y.y.y.y/g" | sed -e "s/\([0-9]\{1,3\}\.\)\{3,3\}[0-9]\{1,3\}/x.x.x.x/g" | sed -e "s/\([0-9a-fA-F]\{2,2\}\:\)\{5,5\}[0-9a-fA-F]\{2,2\}/m.m.m.m/g"
+
+echo
+echo "Docker logs"
```
mayankchhabra, line 97:
```suggestion
echo "Docker containers"
```

scripts/debug
```diff
@@ -0,0 +1,146 @@
...
+    vcgencmd measure_temp
+
+    echo
+    echo "Throttling"
+    echo "----------"
+
+    vcgencmd get_throttled
+
+    echo
+    echo "Startup status"
```
mayankchhabra, line 75:
```suggestion
    echo "umbrel-startup.service logs"
```

scripts/debug
```diff
@@ -0,0 +1,146 @@
...
+    echo "--------------"
+    systemctl status umbrel-startup
```
mayankchhabra, line 77: Since `systemctl` pages the logs, it'd be easier to fetch the last 30 lines via `journalctl`:

```suggestion
    journalctl --unit=umbrel-startup.service | tail -n 30
```

scripts/debug
```diff
@@ -0,0 +1,146 @@
...
+
+ifconfig | sed -e "s/\([0-9a-fA-F]\{1,4\}:\)\{7,7\}[0-9a-fA-F]\{1,4\}/y.y.y.y.y.y.y.y/g" | sed -e "s/[0-9a-fA-F]\{1,4\}:\(:[0-9a-fA-F]\{1,4\}\)\{1,4\}/y::y.y.y.y/g" | sed -e "s/\([0-9]\{1,3\}\.\)\{3,3\}[0-9]\{1,3\}/x.x.x.x/g" | sed -e "s/\([0-9a-fA-F]\{2,2\}\:\)\{5,5\}[0-9a-fA-F]\{2,2\}/m.m.m.m/g"
+
+echo
+echo "Docker logs"
+echo "-----------"
+
+docker-compose logs --tail=100
+
+if [[ ! -z "${UMBREL_OS:-}" ]]; then
```
mayankchhabra, line 102: It's safe to remove this entire `if` block since `umbrel-external-storage.service` already performs these sanity checks for no drive/more than 1 drive and we can spot any errors in its logs.

scripts/debug
```diff
@@ -0,0 +1,146 @@
...
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, 
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, 
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR 
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF 
+# LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING 
+# NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS 
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+# This fork is Copyright: 2020 Umbrel.
+
+[[ -f "/etc/default/umbrel" ]] && source "/etc/default/umbrel"
```
mayankchhabra, line 31: We can drop this since the file is already [sourced in `/etc/profile`](https://github.com/getumbrel/umbrel-os/blob/master/stage2/04-set-umbrel-os-version/00-patches/01-profile.diff#L11).

```suggestion
```

mayankchhabra (changes requested):

scripts/debug
AaronDewes: I added this because it then can give the output of the check and directly tell the user the problem so they don't have to ask.  (Lines 140 - 152)

scripts/debug
AaronDewes: @mayankchhabra Can this then also be removed [in the update scripts](https://github.com/getumbrel/umbrel/blob/master/scripts/update/01-run.sh#L20)?

scripts/debug
mayankchhabra: Ah, okay. Yeah that makes sense!

scripts/debug
mayankchhabra: Update scripts run as sudo and `/etc/profile` isn't automatically sourced for sudo :(

mayankchhabra (approved):
Great job, Aaron! You've reduced tonnes of painful debugging to just a one-liner `ssh -t umbrel@umbrel.local "/home/umbrel/umbrel/scripts/debug"` that we can now pass to anyone facing (almost) any kind of issue with their Umbrel! Awesome work. 👏

Let's for @lukechilds to have a quick look at the script as well before we merge.

scripts/debug
AaronDewes: @mayankchhabra Both users that used the script in the telegram group yet have had the same issue: the script didn't detect Umbrel OS. Should I add it again?

AaronDewes: Maybe you need to add an "export " to the file?

AaronDewes:
A few more improvements are available now, thanks to @louneskmt  for suggesting them.

- The script can now be run with `--upload` to automatically generate a link to share for debugging
- .onion URLs are now no longer in the output
- A few other minor improvements to the result text

louneskmt (approved):

scripts/debug
mayankchhabra: Good spot, Aaron! That is exactly it.

Let's go with sourcing the file for now and then eventually we'll update the `/etc/default/umbrel` file with an export in an OTA update.

scripts/debug
```diff
@@ -0,0 +1,183 @@
...
+cat /proc/swaps
+
+
+echo
+echo "Networking information"
+echo "----------------------"
+echo
+
+ifconfig | sed -e "s/\([0-9a-fA-F]\{1,4\}:\)\{7,7\}[0-9a-fA-F]\{1,4\}/y.y.y.y.y.y.y.y/g" | sed -e "s/[0-9a-fA-F]\{1,4\}:\(:[0-9a-fA-F]\{1,4\}\)\{1,4\}/y::y.y.y.y/g" | sed -e "s/\([0-9]\{1,3\}\.\)\{3,3\}[0-9]\{1,3\}/x.x.x.x/g" | sed -e "s/\([0-9a-fA-F]\{2,2\}\:\)\{5,5\}[0-9a-fA-F]\{2,2\}/m.m.m.m/g"
+
```
mayankchhabra, line 106: We can probably get rid of the networking info as it takes too much space in the logs and we never actually had the need to look at it and instead let's tail `karen's` logs as those are quite useful in the event of a failed update, backup, app installs/uninstalls, etc.

```
tail -n 50 logs/karen.log | sed '/onion/d'
```

mayankchhabra (changes requested):
Really awesome improvements, Aaron! Let's get this in the next release.

scripts/debug
mayankchhabra: Good spot, Aaron! That is exactly it.

Let's go with sourcing the file for now and then eventually we'll update the /etc/default/umbrel file with an export in an OTA update.

AaronDewes:
@mayankchhabra Done. I also added the dmesg output (upload only) because that can help research SSD issues.

lukechilds (approved):
Awesome stuff @AaronDewes, sorry for the delay in reviewing this!

PR #296, opened 2020-10-24
Mention permissions in SECURITY.md

lukechilds:
As we've discussed internally our permissions and usage of root is little too relaxed at the moment and could easily lead to privilege escalation.

We don't currently have this documented in SECURITY.md and it's been pointed out by users viewing the code a few times.


mayankchhabra (approved):
Good to have this documented!

PR #297, opened 2020-10-25
Better support for custom umbrel setups

nioncode:
These are a few small fixes to the configuration scripts that make it easier to run umbrel on custom setups and also improves how the device's ip is detected.

### DIFF
--- a/scripts/configure
+++ b/scripts/configure
@@ -116,7 +116,7 @@ fi
 if [[ -z ${TOR_PASSWORD+x} ]] || [[ -z ${TOR_HASHED_PASSWORD+x} ]]; then
   echo "Generating Tor password"
   echo
-  docker pull --quiet getumbrel/tor:v0.4.1.9
+  docker pull getumbrel/tor:v0.4.1.9
   TOR_PASSWORD=$("./scripts/rpcauth.py" "itdoesntmatter" | tail -1)
   TOR_HASHED_PASSWORD=$(docker run --rm getumbrel/tor:v0.4.1.9 --quiet --hash-password "$TOR_PASSWORD")
 fi
--- a/scripts/configure
+++ b/scripts/configure
@@ -187,7 +187,7 @@ sed -i "s/TOR_PASSWORD=<password>/TOR_PASSWORD=$TOR_PASSWORD/g;" "$ENV_FILE"
 sed -i "s/TOR_HASHED_PASSWORD=<password>/TOR_HASHED_PASSWORD=$TOR_HASHED_PASSWORD/g;" "$ENV_FILE"
 
 # Store docker binary path
-DOCKER_BINARY=$(which docker)
+DOCKER_BINARY=$(readlink -f "$(which docker)")
 sed -i "s#DOCKER_BINARY=<path>#DOCKER_BINARY=$DOCKER_BINARY#g;" "$ENV_FILE"
 
 # Set Umbrel version in electrs banner
--- a/scripts/configure
+++ b/scripts/configure
@@ -58,7 +58,7 @@ if [[ -f "${STATUS_DIR}/configured" ]]; then
 else
   echo "============ CONFIGURING ============="
 fi
-echo "========= UMBREL (${NETWORK}) ==========="
+echo "========= UMBREL (${BITCOIN_NETWORK}) ==========="
 echo "======================================"
 echo
 
--- a/scripts/configure
+++ b/scripts/configure
@@ -186,6 +186,10 @@ sed -i "s/tor.password=<password>/tor.password=$TOR_PASSWORD/g;" "$LND_CONF_FILE
 sed -i "s/TOR_PASSWORD=<password>/TOR_PASSWORD=$TOR_PASSWORD/g;" "$ENV_FILE"
 sed -i "s/TOR_HASHED_PASSWORD=<password>/TOR_HASHED_PASSWORD=$TOR_HASHED_PASSWORD/g;" "$ENV_FILE"
 
+# Store docker binary path
+DOCKER_BINARY=$(which docker)
+sed -i "s#DOCKER_BINARY=<path>#DOCKER_BINARY=$DOCKER_BINARY#g;" "$ENV_FILE"
+
 # Set Umbrel version in electrs banner
 sed -i "/server_banner/s/<version>/$UMBREL_VERSION/g;" "$ELECTRS_CONF_FILE"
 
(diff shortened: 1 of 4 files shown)


AaronDewes (approved):

scripts/start
```diff
@@ -52,7 +52,7 @@ echo "Setting environment variables..."
 echo
 
 # Whitelist device IP, hostname and hidden service for CORS
-DEVICE_IP="$(hostname -I | cut -d ' ' -f 1)"
+DEVICE_IP="$(ip -o route get to 8.8.8.8 | sed -n 's/.*src \([0-9.]\+\).*/\1/p')"
```
lukechilds, line 55: Will this still return the local IP on a local network that doesn't have internet access?

scripts/configure
```diff
@@ -116,7 +116,7 @@ fi
 if [[ -z ${TOR_PASSWORD+x} ]] || [[ -z ${TOR_HASHED_PASSWORD+x} ]]; then
   echo "Generating Tor password"
   echo
-  docker pull --quiet getumbrel/tor:v0.4.1.9
```
lukechilds, line 119: Why did you remove `--quiet` here?

(not arguing it shouldn't be removed, I think it should, just wondering if it related specifically to an issue)

scripts/start
lukechilds: Also, if we're using `ip` now we should probably add it to `check_dependencies`:

https://github.com/getumbrel/umbrel/pull/297/files#diff-ecb3d5615afcbe1e5df55695252c8f0835e8cb1ec10a6048a83cf0a80d2a1ab3R25

scripts/configure
```diff
@@ -58,7 +58,7 @@ if [[ -f "${STATUS_DIR}/configured" ]]; then
 else
   echo "============ CONFIGURING ============="
 fi
-echo "========= UMBREL (${NETWORK}) ==========="
+echo "========= UMBREL (${BITCOIN_NETWORK}) ==========="
```
lukechilds, line 61: So this avoids a bug where empty `NETWORK` var defaults to `BITCOIN_NETWORK=mainnet` but logs empty parentheses here? Is that correct?

If that's the case maybe it would be better to just avoid duplicating `NETWORK`/`BITCOIN_NETWORK` and just do:

```bash
NETWORK="${NETWORK:-mainnet}"
```

Then use `NETWORK` everywhere and remove `BITCOIN_NETWORK`.

scripts/configure
lukechilds: Tbh we can probs nuke this line completely since docker will auto pull it if it doesn't exist.

@mayankchhabra was there a reason the output was suppressed here?

scripts/configure
```diff
@@ -186,6 +186,10 @@ sed -i "s/tor.password=<password>/tor.password=$TOR_PASSWORD/g;" "$LND_CONF_FILE
 sed -i "s/TOR_PASSWORD=<password>/TOR_PASSWORD=$TOR_PASSWORD/g;" "$ENV_FILE"
 sed -i "s/TOR_HASHED_PASSWORD=<password>/TOR_HASHED_PASSWORD=$TOR_HASHED_PASSWORD/g;" "$ENV_FILE"
 
+# Store docker binary path
+DOCKER_BINARY=$(which docker)
```
lukechilds, line 190: If we're trying to autodetect the path here maybe we should resolve any potential symlinks?

```suggestion
DOCKER_BINARY=$(readlink -f "$(which docker)")
```

Lets say `which docker` resolves to `~/bin/docker` which is a symlink to `/some/custom/install/dir/docker`. We'd end up mounting the symlink inside the container which would resolve to `/some/custom/install/dir/docker` inside the container and break.

lukechilds (changes requested):
Great stuff!

Couple of minor nits but great contribution. Thanks a lot @nioncode!

scripts/configure
nioncode: The problem was just that `${NETWORK}` was never assigned and I got an error from my shell. To me it seems like `NETWORK` is only used in the configure and start scripts and all other files like `docker-compose` use `BITCOIN_NETWORK`. So if we want to consolidate `NETWORK` and `BITCOIN_NETWORK`, I'd suggest to drop `NETWORK` and use `BITCOIN_NETWORK` everywhere.

scripts/configure
nioncode: Yes, as specified in the commit message for this change, `--quiet` option is not supported in earlier docker versions and thus the configure script fails.
I'd be in favor of removing the complete docker pull.

scripts/start
nioncode: Thanks, I added `ip` to the dependency check.

The method does not require internet access. The ip that is given in the command is never contacted, i.e. no network request is made. It just asks the system over which interface the request WOULD be routed if it is made, which should give you the IP address of your primary interface.

lukechilds: Awesome TIL!

scripts/configure
lukechilds: Ah, makes sense!

Sorry, totally missed your commit descriptions, reading though them now!

scripts/configure
lukechilds: Sounds good to me.

lukechilds: I agree `BITCOIN_NETWORK ` is a better name but it's actually not simple to change.

It breaks anything external that relies on it, `umbrel-dev` for example: https://github.com/getumbrel/umbrel-dev/blob/21ac2417e687b219ef4ca629d526ca19f2660eab/Vagrantfile#L64

If we change the name we break all umbrel-dev installs. Even if we update umbrel-dev we need to tell users to upgrade and we need to specify both versions of the var incase users are developing on a branch forked from an old version of Umbrel.

Plus it breaks any existing guides/scripts/etc.

Ok, lets stick with your current solution.

scripts/configure
lukechilds: Yeah, lets nuke it.

mayankchhabra: > @mayankchhabra was there a reason the output was suppressed here?

I don't think there was any reason, except making the logs cleaner. We also suppress outputs while doing a `docker-compose pull` later in the configure script:

https://github.com/getumbrel/umbrel/blob/b748b7bc331f55a23e6cc5d46b4f550d0af45dec/scripts/configure#L236

Shall we also remove that too then?

nioncode: `docker-compose` did support the `--quiet` flag on my system, so I did not remove it. Whether it makes sense to reduce the output information here (and thus keep `--quiet`) is up to you guys.

lukechilds:
Thanks a lot for this @nioncode!

Just FYI, we have a pretty strict commit policy of 1 PR = 1 atomic change = 1 final squashed commit in master.

Our fault, we haven't specified this anywhere in the readme, and this pretty much started as a single change and then grew into a few other things based on feedback.

I've pulled the individual changes out into separate commits in master and referenced this PR.

Great contribution!

PR #301, opened 2020-10-28
Add how-to upgrade Umbrel from Command Line (when this is preferred over OTA updates)

maltokyo:
Just READMe update

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,20 +65,11 @@ To stop Umbrel, run:
 sudo ./scripts/stop
 ```
 
-## 🎹 Services orchestrated
+### Optional: How-to Upgrade Umbrel from the command line**
 
-- [`Umbrel Dashboard`](https://github.com/getumbrel/umbrel-dashboard)
-- [`Umbrel Manager`](https://github.com/getumbrel/umbrel-manager)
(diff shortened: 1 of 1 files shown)


louneskmt (changes requested):
You will need to change your markdown. As you can [see](https://github.com/getumbrel/umbrel/blob/9267b7c80f930e853ec0e2f3a89870237890cf5d/README.md), the style needs to be improved. Please adopt the same one as the original README.

maltokyo:
Thank you @louneskmt - done.  https://github.com/getumbrel/umbrel/blob/ffb25fa948e1f5abea6010fa8bb1ea22fa5f1506/README.md

louneskmt (commented):
Much better, now let's wait to see if Mayank or Luke approve it 😁

lukechilds:
Appreciate you taking the time to add this @maltokyo but unfortunately this is not at all safe to do and not something we want to recommend.

There is no official way to upgrade via CLI. You need to upgrade via the UI.

It's possible to use the `update` commands to upgrade via CLI, however this is intended for development use only and is completely unsupported. It bypasses many safety checks and could result in a broken system.

See here for more details: https://github.com/getumbrel/umbrel/issues/205#issuecomment-691475831

PR #312, opened 2020-11-12
Umbrel v0.2.14

lukechilds:
This Umbrel release brings LND v0.11.1, native segwit addresses for on-chain deposits and a few tiny but mighty changes that will silently prepare your Umbrel for some big upcoming updates.

Diff: https://github.com/getumbrel/umbrel/compare/v0.2.13...v0.2.14


mayankchhabra (approved):
Let's gooooooooo

PR #318, opened 2020-11-20
Fix lnd refreshing TLS certificate at each restart

louneskmt:
This PR adds a setting to not include the container hostname (its Docker container ID) in the certificate. Thus, lnd won't detect changes at each restart and won't create a new certificate, which was leading to invalid connection on Lightning wallets like Zap after a restart.


lukechilds (approved):

lukechilds:
Great stuff, thanks @louneskmt!

PR #320, opened 2020-11-25
Add FAQ

AaronDewes:
This are the FAQ that I already shared in Telegram with a few improvements. This can possibly avoid lots of questions on Telegram. @mayankchhabra I think you are developing the website, in my opinion it should also be added there, many people don't check the GitHub repo before asking questions on Telegram (for example with the SSH password). 
It's probably possible to improve the grammar, I'm not so good at English.


mayankchhabra (changes requested):
Thank you, @AaronDewes! FAQs really are the need of the hour. I will add them to the website once we have refined them a bit.

Meanwhile, can we please move them to README.md instead (above the contributing section)? To keep things clean, we can have them as a collapsible list:

<details>
    <summary>Like this</summary>

  Here's the [markdown](https://gist.github.com/pierrejoubert73/902cc94d79424356a8d20be2b382e1ab)
</details>

pors:
I'd like to add an FAQ, or possibly more than one. Where/how can I do that? Here in the comments? Or in the README.md?

AaronDewes:
Do it here in the comments.

pors:
Q: The wallet doesn't seem to work from the command line:

```
$ ~/umbrel/bin/bitcoin-cli getbalance
0.00000000
```

A: Umbrel doesn't use the bitcoin-core wallet but the one provided by LND:

```
$ ~/umbrel/bin/lncli walletbalance
{
    "total_balance": "95003",
    "confirmed_balance": "95003",
    "unconfirmed_balance": "0"
}
```

AaronDewes:
> Q: The wallet doesn't seem to work from the command line:
> 
> ```
> $ ~/umbrel/bin/bitcoin-cli getbalance
> 0.00000000
> ```
> 
> A: Umbrel doesn't use the bitcoin-core wallet but the one provided by LND:
> 
> ```
> $ ~/umbrel/bin/lncli walletbalance
> {
>     "total_balance": "95003",
>     "confirmed_balance": "95003",
>     "unconfirmed_balance": "0"
> }
> ```

Sorry for the late reply. In my opinion, this FAQ is for new users and for the general users of Umbrel, and users aren't supposed to access the CLI on Umbrel, that's why in my opinion, we don't need to document them, butter better wait for a replay from Luke or Mayank.

PR #332, opened 2020-12-10
support custom DNS resolvers

nioncode:
At the moment, umbrel expects that a) you have a mDNS daemon (e.g. avahi) on the machine that runs umbrel (which is a reasonable assumption for umbrel-os, but not for custom installs) and b) that the client accessing umbrel has a mDNS resolver (some Linux distros don't have one by default). If you don't fulfill both requirements, you cannot access `umbrel.local`, since it will not resolve.

Some routers seem to implement a custom DNS resolver and send this to all clients in order to do domain lookups. This is e.g. the case for fritzbox routers (which are very popular in Germany). With these routers, you can simply resolve any hostname in the local network by name (e.g. `http://umbrel.fritz.box/`) and you can even omit the trailing domain, i.e. `http://umbrel/` will resolve just fine as well. This is achieved by some kind of handshake between the router and the host system during dhcp, which updates `/etc/resolv.conf` to contain the domain search path, e.g.:
```
# Generated by resolvconf
search fritz.box
nameserver 192.168.1.1
nameserver 1.1.1.1
nameserver 1.0.0.1
options edns0
```
or 
```
#PROTO: DHCP
domain fritz.box
nameserver 192.168.1.1
bootserver 192.168.1.1
```

In order to support to connect to umbrel via these custom domain suffixes, they must be added to the CORS whitelist, which is done with this PR. If no custom domain search path was found it `/etc/resolv.conf`, this PR falls back to `.local` as before, so setups without a custom DNS suffix should not be affected.

One thing that I just thought of was that we might still want to add the `.local` url to the CORS whitelist as well, just to make sure that you can use both, your custom domain name or `.local`, if your network supports both (which I have not tested). Please let me know if this is more reasonable, then I'll try to integrate that.

@lukechilds we discussed this a few weeks back in telegram, but somehow I never put up the PR for this. I'm using this in my local network since a few weeks without problems.

### DIFF
--- a/scripts/start
+++ b/scripts/start
@@ -30,6 +30,7 @@ check_dependencies rsync jq curl
 
 UMBREL_ROOT="$(dirname $(readlink -f "${BASH_SOURCE[0]}"))/.."
 UMBREL_LOGS="${UMBREL_ROOT}/logs"
+ENV_FILE="${UMBREL_ROOT}/.env"
 
 if [[ ! -d "$UMBREL_ROOT" ]]; then
   echo "Root dir does not exist '$UMBREL_ROOT'"
@@ -51,16 +52,19 @@ echo
 echo "Setting environment variables..."
 echo
 
+# Source persisted environment variables.
+. $ENV_FILE
+
 # Whitelist device IP, hostname and hidden service for CORS
 DEVICE_IP="$(ip -o route get to 8.8.8.8 | sed -n 's/.*src \([0-9.]\+\).*/\1/p')"
 DEVICE_HOSTNAME="$(hostname)"
-DEVICE_HOSTS="http://${DEVICE_IP},http://${DEVICE_HOSTNAME}.local,https://${DEVICE_HOSTNAME}.local,http://${DEVICE_HOSTNAME},https://${DEVICE_HOSTNAME}"
+DEVICE_HOSTS="http://${DEVICE_IP},http://${DEVICE_HOSTNAME}.local,https://${DEVICE_HOSTNAME}.local,http://${DEVICE_HOSTNAME},https://${DEVICE_HOSTNAME},http://${DEVICE_HOSTNAME}${DEVICE_DNS_SUFFIX},https://${DEVICE_HOSTNAME}${DEVICE_DNS_SUFFIX}"
 if [[ -f "${UMBREL_ROOT}/tor/data/web/hostname" ]]; then
     hidden_service_url=$(cat "${UMBREL_ROOT}/tor/data/web/hostname")
     DEVICE_HOSTS="${DEVICE_HOSTS},http://${hidden_service_url}"
 fi
 export DEVICE_HOSTS=$DEVICE_HOSTS
-export DEVICE_HOSTNAME="${DEVICE_HOSTNAME}.local"
+export DEVICE_HOSTNAME="${DEVICE_HOSTNAME}${DEVICE_DNS_SUFFIX}"
 
 # Increase default Docker and Compose timeouts to 240s
 # as bitcoin can take a long while to respond
--- a/scripts/configure
+++ b/scripts/configure
@@ -193,8 +193,27 @@ sed -i "s#DOCKER_BINARY=<path>#DOCKER_BINARY=$DOCKER_BINARY#g;" "$ENV_FILE"
 sed -i "/server_banner/s/<version>/$UMBREL_VERSION/g;" "$ELECTRS_CONF_FILE"
 
 # Add hostname to lnd.conf for TLS certificate
+RESOLV_CONF="/etc/resolv.conf"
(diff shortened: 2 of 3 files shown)


AaronDewes (approved):

AaronDewes (changes requested):
> One thing that I just thought of was that we might still want to add the .local url to the CORS whitelist as well, just to make sure that you can use both, your custom domain name or .local, if your network supports both (which I have not tested). Please let me know if this is more reasonable, then I'll try to integrate that.

On my FRITZ!Box, I'm using umbrel.local without problems, and I think breaking support for umbrel.local would be very confusing for most users, so in my opinion you should also allow umbrel.local

```suggestion
DEVICE_HOSTS="http://${DEVICE_IP},http://${DEVICE_HOSTNAME}${DEVICE_DNS_SUFFIX},https://${DEVICE_HOSTNAME}${DEVICE_DNS_SUFFIX},http://${DEVICE_HOSTNAME},https://${DEVICE_HOSTNAME},http://${DEVICE_HOSTNAME}.local,https://${DEVICE_HOSTNAME}.local"
```

Doing it like that should work.

scripts/start
```diff
@@ -51,16 +52,19 @@ echo
...
 echo
 
+# Source persisted environment variables.
+. $ENV_FILE
+
 # Whitelist device IP, hostname and hidden service for CORS
 DEVICE_IP="$(ip -o route get to 8.8.8.8 | sed -n 's/.*src \([0-9.]\+\).*/\1/p')"
 DEVICE_HOSTNAME="$(hostname)"
-DEVICE_HOSTS="http://${DEVICE_IP},http://${DEVICE_HOSTNAME}.local,https://${DEVICE_HOSTNAME}.local,http://${DEVICE_HOSTNAME},https://${DEVICE_HOSTNAME}"
+DEVICE_HOSTS="http://${DEVICE_IP},http://${DEVICE_HOSTNAME}${DEVICE_DNS_SUFFIX},https://${DEVICE_HOSTNAME}${DEVICE_DNS_SUFFIX},http://${DEVICE_HOSTNAME},https://${DEVICE_HOSTNAME}"
```
AaronDewes, line 61:
```suggestion
DEVICE_HOSTS="http://${DEVICE_IP},http://${DEVICE_HOSTNAME}${DEVICE_DNS_SUFFIX},https://${DEVICE_HOSTNAME}${DEVICE_DNS_SUFFIX},http://${DEVICE_HOSTNAME},https://${DEVICE_HOSTNAME},http://${DEVICE_HOSTNAME}.local,https://${DEVICE_HOSTNAME}.local"
```

Doing it like that should work.

AaronDewes (changes requested):

nioncode:
> On my FRITZ!Box, I'm using umbrel.local without problems, and I think breaking support for umbrel.local would be very confusing for most users, so in my opinion you should also allow umbrel.local.

Ah yes, this should of course not be broken in that scenario then. I'll try to integrate this.

scripts/start
nioncode: Is something like this also needed in lnd's `tlsextradomain` config?

AaronDewes: I'm not sure, but I think yes

nioncode: Principally, it is not required in `tlsextradomain`, since there the value of `DEVICE_HOSTNAME` should be used, which now is the hostname with the custom DNS suffix, so these are in sync. HOWEVER, since this will affect the LND connect URLs, these will change for users on the first start of the new version, which will break their existing lightning wallet connections, which is bad.

The only solution to this that I can think of is trying to resolve hostname.local in scripts/configure. If this works, use hostname.local everywhere and ignore the custom DNS suffix. This provides a smooth update path for existing users.

Simply adding another tlsextradomain entry to LND's config will make LND accept these connections, but it will change the TLS certificate during the update and thus existing connections will be broken anyways and must be re-established. Thus adding the extra domain does not really help.

nioncode: Unfortunately I could not solve this problem sufficiently. When I ping my umbrel node at `umbrel.local` from the box itself, it works fine (because it has `avahi` installed), but my target machine does not have a mDNS resolver, so it cannot connect.

In the end I got avahi installed on my target machine and can now connect on `umbrel.local`, so I guess this PR is most likely not needed anymore, since the recommendation is to either get a mDNS resolver installed or connect to umbrel through the local IP address instead to simplify support requests.

@AaronDewes what do you think?

AaronDewes: Oh sorry, forgot to respond, I don't think this PR is needed anymore, but @lukechilds and @mayankchhabra schould decide that.

nioncode:
Closing as per https://github.com/getumbrel/umbrel/pull/332#discussion_r542876113. The recommendation is to either install avahi on the machine connecting to umbrel or use the IP address instead.

PR #333, opened 2020-12-13
Add app framework

lukechilds:
This PR adds the preliminary version of the app framework.

Third party apps can add a package at `$UMBREL_ROOT/apps/$APP_ID/` that can be dynamically installed/uninstalled by Umbrel users at runtime.

Apps are implemented as Docker Compose configurations, they can persist their own application storage that will be removed at uninstall.

The long term goal of the app framework is to have apps live external to the main Umbrel repo. They will be able to be updated by the developers independently of Umbrel and apps will each run on an isolated network and have to register fine grained permissions to get access to different services.

However for now, each app lives inside the Umbrel repo connects directly to the main Umbrel Docker network with access to every service. We then manually review the configurations before merging the PRs.

Further documentation here: https://github.com/getumbrel/umbrel/blob/265257da5ab6fc7b487a19b9b449b5475b6957e3/apps/README.md


mayankchhabra (approved):
YES

image

mayankchhabra:
#333 — this one will be long remembered!

louneskmt:
Congrats eChads 🎉

PR #343, opened 2020-12-17
Add thunderhub app

apotdevin:
# App Submission

### App name
ThunderHub

### Version
v0.12.1

A simple and intuitive UI to take full control of your Lightning Node

### Summary of the app
ThunderHub allows you to take full control of your Lightning Node with a slick and awesome UI. Explore all the options that ThunderHub has to offer, from sending and receiving Lightning payments, checking your nodes health statistics and even more advanced options like channel rebalancing and multi-path lightning payments. Managing and monitoring your node has never been easier with transaction reports, graphs and a huge assortment of different features all bundled inside of this great tool.

### Developed by
@apotdevin

### Developer website
https://apotdevin.com/

### Source code repository
https://github.com/apotdevin/thunderhub

### Support link
https://t.me/thunderhub

### Uses
- [ ] Bitcoin Core
- [ ] Electrum server
- [X] LND

### 256x256 SVG icon
ThunderHub

### App screenshots
NightLight
Forwards
Channels

### I have tested my app on:
- [x] [Umbrel dev environment](https://github.com/getumbrel/umbrel-dev)


apps/thunderhub/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+      - ${LND_DATA_DIR}:/lnd:ro
+    environment:
+      NO_VERSION_CHECK: "true"
+      NO_CLIENT_ACCOUNTS: "true"
+      LOG_LEVEL: "debug"
+      SSO_SERVER_URL: "https://${LND_IP}:${LND_GRPC_PORT}"
+      SSO_MACAROON_PATH: "${LND_DATA_DIR}/data/chain/bitcoin/${BITCOIN_NETWORK}"
+      SSO_CERT_PATH: "${LND_DATA_DIR}/tls.cert"
+      DANGEROUS_NO_SSO_AUTH: "true"
+      LOGOUT_URL: "http://umbrel-dev.local"
```
AaronDewes, line 27: You should add compatibility with the current Umbrel URL.
```suggestion
      LOGOUT_URL: ${DEVICE_HOSTS:-"http://umbrel.local"}
```

apps/thunderhub/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+      - 3000:3000
+    volumes:
+      - ${LND_DATA_DIR}:/lnd:ro
+    environment:
+      NO_VERSION_CHECK: "true"
+      NO_CLIENT_ACCOUNTS: "true"
+      LOG_LEVEL: "debug"
+      SSO_SERVER_URL: "https://${LND_IP}:${LND_GRPC_PORT}"
+      SSO_MACAROON_PATH: "${LND_DATA_DIR}/data/chain/bitcoin/${BITCOIN_NETWORK}"
+      SSO_CERT_PATH: "${LND_DATA_DIR}/tls.cert"
```
AaronDewes, line 25: It accesses the mount point inside the container, so you can hardcode these paths.
```suggestion
      SSO_MACAROON_PATH: "/lnd/data/chain/bitcoin/${BITCOIN_NETWORK}"
      SSO_CERT_PATH: "/lnd/tls.cert"
```

apps/thunderhub/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
+version: "3.7"
+
+x-logging:
+  &default-logging
+  driver: journald
+  options:
+    tag: "umbrel-app {{.Name}}"
+
+services:
+  thub:
```
AaronDewes, line 10: This should always be called `web` (was pointed out in the developer chat)
```suggestion
  web:
```

apps/thunderhub/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+    image: apotdevin/thunderhub:v0.12.0
```
AaronDewes, line 11: This should be pinned to a specific sha-256

AaronDewes (changes requested):

AaronDewes (dismissed):

apps/thunderhub/docker-compose.yml
mayankchhabra: `$DEVICE_HOSTS` is an array of all the possible URLs and IP addresses that the user could be accessing their Umbrel at. It wouldn't exactly tell you which URL the user is currently at (umbrel.local, Tor hidden service, or IP address) plus we don't expose this variable to apps (as it can leak the dashboard's hidden service address).

It would perhaps be better to drop the redirect logic entirely, 'cause even though you can get the IP address/hostname via `window.hostname` in the browser on local network to correctly figure out the right redirect URL, you won't be able to figure out dashboard's onion address if the user is accessing via Tor.

apps/thunderhub/docker-compose.yml
mayankchhabra: While the image should be pinned to the SHA-256 digest of it's multi-arch manifest list, we don't expect the app developers to do that. We'll pin it ourselves as a part of the final changes before merge:

> After you've submitted your app, we'll review your pull request, create the required Tor hidden services for it, make some adjustments in the docker-compose.yml file, such as removing any port conflicts with other apps, pinning Docker images to their sha256 digests, assigning unique IP addresses to the containers, etc before merging. 
>
> https://github.com/getumbrel/umbrel/blob/master/apps/README.md

apps/thunderhub/data/thubConfig.yaml
```diff
@@ -0,0 +1,14 @@
+masterPassword: 'moneyprintergobrrr'
+accounts:
+  - name: 'Umbrel'
+    serverUrl: '10.0.0.9:10009'
```
louneskmt, line 4: Hey! I think you should use the env variable LND_IP et LND_PORT instead of hard coding the host:port. You can for example set a container env variable for the server url:
```
environment:
    - SERVER_URL: $LND_IP:$LND_PORT
```
```suggestion
    serverUrl: '${SERVER_URL}'
```
Or just set the same env var and use
```suggestion
    serverUrl: '${LND_IP}:${LND_PORT}'
```

apotdevin: Hey! Yeah plan on adding the functionality to read env vars inside the config file in a next release but it's currently not possible so had to hard code them

lukechilds (approved):
Awesome work @apotdevin!

Tested and working on both umbrel-dev and my mainnet Umbrel.

I've wired up the Tor hidden service and pinned the Docker Hub image.

I also changed the container to run as UID 1000 since it was running as root.

Those hardcoded connection details really do need to be set dynamically at some point too. It'll break if we ever change the container IPs which we likely will. The plan is that they will be assigned dynamically soon. But no need to let that block merging this.

Great stuff!

apotdevin:
Fixed the hardcoded value in the yaml config file 🥳

lukechilds (approved):
Re-tested the latest changes and pinned the Docker image.

Thanks for resolving that so quickly!

mayankchhabra (approved):
This is just incredible work, @apotdevin! Tried it out on my testnet Umbrel and it works flawlessly on both local network and Tor. 

image

lukechilds:
Merged!

Congrats on being the first third-party app integrated into Umbrel @apotdevin!

PR #345, opened 2020-12-18
Reduce dbcache post IBD

lukechilds:
By default we set `dbcache` to 50% total ram - 300MB since it helps significantly speed up IBD.

300MB is the default mempool size so we end up allocating ~50% of total system memory to Bitcoin Core.

However this is incredibly wasteful once IBD has completed, especially on high memory systems, a huge amount of RAM is being allocated to Bitcoin Core when it really isn't needed.

This PR reduces `dbcache` to 200MB once IBD has completed. That means post IBD 200MB `dbcache` and 300MB `maxmempool` are allocated to Bitcoin Core for a total of ~500MB.

This should be plenty since `dbcache` is only used to cache the UTXO set to speed up block validation. Post IBD we only need to validate one block every ~10 minutes so it's not particularly time critical.

We can probably reduce this value even further, the minimum allowed value for `dbcache` is 4. 200 seems sensible for now but we should test reducing right down to 4.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -155,12 +155,13 @@ services:
                         ipv4_address: $MIDDLEWARE_IP
         neutrino-switcher:
                 container_name: neutrino-switcher
-                image: getumbrel/neutrino-switcher:v1.0.3
+                image: getumbrel/neutrino-switcher:v1.1.1
                 logging: *default-logging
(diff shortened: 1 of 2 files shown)


mayankchhabra (approved):
Looks great! Nice work, man. 4mb `dbache` would be sweet!

PR #346, opened 2020-12-18
Remove "

AaronDewes:
This makes everything in the file have the same style. In addition, Umbrel didn't work for me until I removed the ".


mayankchhabra (approved):
Thanks, @AaronDewes! Looks good to me. Curious to know what was the error that you faced with quotes?

AaronDewes:
I got this for the network configuration of the whole compose file:
> invalid CIDR address

for the IP addresses
> invalid IPv4 address

PR #348, opened 2020-12-18
Add lightning-terminal app

louneskmt:
# App Submission

### App name
Lightning Terminal

### Version
v0.3.4-alpha

_(max 50 characters)_

The easiest way to manage channel liquidity.

### Summary of the app
_(50 to 200 words)_

Lightning Terminal is the easiest way to manage inbound and outbound liquidity on the Lightning Network. Keep your channels open and the funds flowing. It provides a visual interface for interacting with your channels and balances using Loop. Making the process of managing your funds easier and more intuitive allows you to focus on running your business, not on your liquidity. Loop is the first of several upcoming Terminal features that enable new use cases and open up a new world of possibilities on Lightning.

Why use Loop?
- Add "inbound liquidity" to receive payments
- Reduce transaction fees by recycling and reusing Lightning channels
- Send funds to and from users or services that aren't yet Lightning enabled
- Configurable wait times and "batching" allow for further fee savings
- Refill and offload funds from any number of Lightning channels in a single on-chain transaction

Pricing:
- Loop Out: 0.05-0.5%
- Loop In: 0.10-1%
- Loop pricing varies based on wait time and on-chain fee conditions.

### Developed by
Lightning Labs (@lightninglabs)

### Developer website
https://lightning.engineering

### Source code repository
https://github.com/lightninglabs/lightning-terminal

### Support link
_(Link to your Telegram support channel, GitHub issues/discussions, support portal, or any other place where users could contact you for support.)_
https://github.com/lightninglabs/lightning-terminal/issues/new

### Uses
- [ ] Bitcoin Core
- [ ] Electrum server
- [x] LND

### 256x256 SVG icon
_(GitHub doesn't allow uploadig SVGs directly. Upload your file to an alternate service, like https://svgur.com, and paste the link below.)_

icon

### App screenshots
_(Upload 3 to 5 high-quality screenshots (at least 1280x800px) of your app in PNG format.)_

image
image
image

### I have tested my app on:
- [x] [Umbrel dev environment](https://github.com/getumbrel/umbrel-dev)


louneskmt:
This PR depends on https://github.com/lightninglabs/lightning-terminal/pull/163. It will not work without it being merged as this last introduces a new option allowing the use of an insecure HTTP server instead of HTTPS. The image version will be updated when the PR is merged.

lukechilds (approved):
Awesome stuff @louneskmt!

Made a few tweaks so the container can run as non-root and also a few datadirs were not being persisted so added them in.

lukechilds:
Also @jamaljsr @guggero would be great to have your 2c on this and let us know if everything looks ok.

apps/lightning-terminal/docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+        - --lit-dir="/lit"
+        - --faraday.faradaydir="/faraday"
+        - --loop.loopdir="/loop"
+        - --pool.basedir="/pool"
+        - --insecure-httplisten=0.0.0.0:$APP_LIGHTNING_TERMINAL_PORT
+        - --lnd-mode="remote"
+        - --remote.lnd.network="$BITCOIN_NETWORK"
+        - --remote.lnd.rpcserver=$LND_IP:$LND_GRPC_PORT
+        - --remote.lnd.macaroondir="/lnd/data/chain/bitcoin/$BITCOIN_NETWORK"
+        - --remote.lnd.tlscertpath="/lnd/tls.cert"
```
guggero, line 34: This whole section could be simplified somewhat by storing everything in the same directory:
```
    restart: on-failure
    stop_grace_period: 5m
    ports:
      - "$APP_LIGHTNING_TERMINAL_PORT:$APP_LIGHTNING_TERMINAL_PORT"
    environment:
      - HOME=/home/lit
    volumes:
      - ${APP_DATA_DIR}/lit:/home/lit
      - ${LND_DATA_DIR}:/lnd:ro
    command:
        - --insecure-httplisten=0.0.0.0:$APP_LIGHTNING_TERMINAL_PORT
        - --lnd-mode="remote"
        - --remote.lnd.network="$BITCOIN_NETWORK"
        - --remote.lnd.rpcserver=$LND_IP:$LND_GRPC_PORT
        - --remote.lnd.macaroondir="/lnd/data/chain/bitcoin/$BITCOIN_NETWORK"
        - --remote.lnd.tlscertpath="/lnd/tls.cert"
```

This would mean that the config file in Umbrel should be moved to `apps/lightning-terminal/lit/.lit/lit.conf`.
Of course this is just a suggestion and not necessary. I just noticed that the directory flags are called differently for each project (`loopdir` vs. `basedir` vs. `faradaydir` :grimacing: ) and this would make it a bit more future proof if we ever decide to rename them.

guggero (dismissed):
Awesome work on this! Looks good to me, just a small suggestion for a simplification (assuming it doesn't make it more difficult with respect to directory ownerships).

apps/lightning-terminal/docker-compose.yml
lukechilds: Ahh yes great spot!

Also mean we don't need to go creating new dirs and adding new flags if you bundle new daemons into LiT.

mayankchhabra (approved):
Really great job on packaging LiT, @louneskmt! Tested out on mainnet and it's working flawlessly on both local network and Tor. Haven't tried loop in or out yet but will do that as a part of our E2E testing of all apps before the upcoming release.

The interface is super clean, love it! Awesome work, guys @jamaljsr @guggero! Please also feel free to suggest edits to the PR description if you'd like. All of that information is going to be user-facing in a dedicated UI.

jamaljsr:
Really awesome job everyone! I'm super excited to see Terminal package with Umbrel.

My only 2 cents to add would be regarding the password for Terminal. I'm a little bit uneasy using a static password for all Umbrel users. `litd` has a `--ui-password_env=<ENVVAR>` flag which can be used instead of storing the current value of `moneyprintergobrrr` in the `lit.conf` file. 

Is there any way to pass down user defined config values from Umbrel to the apps via ENV vars? I saw that there is `$BITCOIN_RPC_PASS` available in the compose file, but I'm not sure if this is auto-generated, static, or user-defined. I'm not sure if there's any other method to make the password user-specific.

I wouldn't consider this blocking for the initial launch, but should be something we should revisit in a future release.

lukechilds:
@jamaljsr 

>My only 2 cents to add would be regarding the password for Terminal. I'm a little bit uneasy using a static password for all Umbrel users.

We've had quite a few discussions about this internally. Basically the auth is pretty much redundant and just there as a UX thing to make the user feel better.

On the local network we can't generate CA signed SSL certs and most users are not technically competent enough to correctly configure self signed certs so all local communication is in plain text. If there is someone malicious on the local network they can just sniff the password. Allowing users to set their own password could be a potential foot gun if they reuse a password they use elsewhere and are unknowingly broadcasting it in plain text over the local network.

Also Umbrel is currently in beta, we're currently working on the assumption that the local network is secure, so it doesn't make sense to worry too much about local connections at this point.

When accessing remotely over Tor the hidden service is providing 256 bits of entropy. No matter how weak or strong the password is, or even if there was no password at all, an attacker first need to brute force the hidden service which is not possible.

Long term, before a stable release, we do plan to implement proper auth. This helps add an additional layer of security in the event that a user accidentally leaks their hidden service (e.g a Tor Browser screenshot posted on Twitter). However this is not something that will need to be implemented at the app level. We can implement SSO at the Umbrel level by running an auth proxy in front of every app container. But that's not something we need to worry about for now.

Thanks for raising the concern though. Let me know if you disagree with any of our reasoning.

jamaljsr:
@lukechilds  You have clearly thought about this far more than I have, and have brought up some good points in regards to SSL and Tor security that I hadn't thought of. I think your assumptions are fair. Thanks for the detailed response. 👍

PR #360, opened 2020-12-26
Kill apps (with kindness) for faster uninstalls

mayankchhabra:


mayankchhabra:
Bad idea.

> The app could be communicating with a remote server and it might be important to shut down cleanly
E.g if loop was mid swap and we just kill the process

h/t @lukechilds 

Makes more sense to just reduce the shutdown grace period of all apps to `1m` instead.

PR #371, opened 2020-12-29
Create app directory before pulling images

louneskmt:
Create the app directory and copy app data before reading the compose file for the first time. For instance, if the compose file uses a local file, it won't find it.

Example :
```
# apps/<app>/docker-compose.yml
...
env_file:
    - ${APP_DATA_DIR}/app.conf
```
currently results in 
```
Pulling images for app <app>...
ERROR: Couldn't find env file: /home/umbrel/umbrel/app-data/<app>/app.conf
```

@lukechilds

### DIFF
--- a/scripts/app
+++ b/scripts/app
@@ -109,12 +109,12 @@ update_installed_apps() {
 # Pulls down images for an app and starts it
 if [[ "$command" = "install" ]]; then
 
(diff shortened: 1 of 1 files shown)


mayankchhabra (approved):
Very good find, @louneskmt! Thanks for the fix :)

PR #378, opened 2020-12-30
Fix sphinx relay version

AaronDewes:


lukechilds (approved):
Good spot, thanks Aaron!

Issue #291, opened 2020-10-17
License issues


AaronDewes:
> For the avoidance of doubt, You do not have permission under this Public License to Share Adapted Material.

If I understand this part correctly, the NoDerivatives version
disallows contributing to Umbrel, because by submitting a PR, I share a modified version of Umbrel publicly.

mayankchhabra:
Unfortunately that's one downside of using a readymade license. It's on our list to do a custom license that does allow derivate work for upstream contributions. For now, we're happy to provide explicit written consent to every developer making contributions (on request).

mayankchhabra:
Closing this for now as it's a long-term to-do. Meanwhile @AaronDewes if you'd like an explicit written consent, feel free to hit me up directly.

AaronDewes:
I wrote a license you could use, it is adapted from the Licnese you're using currently, available under CC0 and provided without warranty or liability. It hasn't been checked by a lawyer, and there might be legal loopholes. 



## Aaron Dewes Public License v0.1

By exercising the Licensed Rights (defined below), You accept and agree to be bound by the terms and conditions of this Aaron Dewes License v0.1 ("Public License"). To the extent this Public License may be interpreted as a contract, You are granted the Licensed Rights in consideration of Your acceptance of these terms and conditions, and the Licensor grants You such rights in consideration of benefits the Licensor receives from making the Licensed Material available under these terms and conditions.

### Section 1 – Definitions.

a. __Adapted Material__ means material subject to Copyright and Similar Rights that is derived from or based upon the Licensed Material and in which the Licensed Material is translated, altered, arranged, transformed, or otherwise modified in a manner requiring permission under the Copyright and Similar Rights held by the Licensor. For purposes of this Public License, where the Licensed Material is a musical work, performance, or sound recording, Adapted Material is always produced where the Licensed Material is synched in timed relation with a moving image.

b. __Copyright and Similar Rights__ means copyright and/or similar rights closely related to copyright including, without limitation, performance, broadcast, sound recording, and Sui Generis Database Rights, without regard to how the rights are labeled or categorized. For purposes of this Public License, the rights specified in Section 2(b)(1)-(2) are not Copyright and Similar Rights.

e. __Effective Technological Measures__ means those measures that, in the absence of proper authority, may not be circumvented under laws fulfilling obligations under Article 11 of the WIPO Copyright Treaty adopted on December 20, 1996, and/or similar international agreements.

[...]

PR #385, opened 2021-01-01
Submit squeaknode app

yzernik:
# App Submission

### App name
squeaknode

### Version
0.1.5

_(max 50 characters)_

Squeaknode is a peer-to-peer status feed.

### Summary of the app
_(50 to 200 words)_

The squeaknode app allows you to create, view, buy, and sell squeaks. A squeak is a single post that can contain up to 280 utf-8 characters. 
The squeaknode timeline is ordered by the timestamp/blockhash embedded in each squeak.
Each squeak is signed with a digital signature of the private key of the author.
Squeaks can be downloaded from any peer to any peer, but they remain locked until the downloader makes a lightning payment to decrypt the content.

### Developed by
Jonathan Zernik

### Developer website
https://github.com/yzernik

### Source code repository
https://github.com/yzernik/squeaknode

### Support link
https://github.com/yzernik/squeaknode/discussions

### Uses
- [x] Bitcoin Core
- [ ] Electrum server
- [x] LND

### 256x256 SVG icon
https://svgshare.com/s/SmD

### App screenshots
Screenshot from 2021-01-29 21-49-17
Screenshot from 2021-01-29 21-50-12
Screenshot from 2021-01-29 21-50-43
Screenshot from 2021-01-29 21-51-37
Screenshot from 2021-01-29 21-52-01

### I have tested my app on:


mayankchhabra:
Wow, that's a really very interesting app @yzernik! If my understanding is correct, is Squeaknode like a peer-to-peer twitter where you can buy (unlock) and sell tweets with micropayments on the Lightning Network?

yzernik:
@mayankchhabra Yes that is a good summary. The long term goal is that the lightning payment will trustlessly unlock the content, but that is not possible until the next version of LND supports PTLC channels.

apps/squeaknode/docker-compose.yml
```diff
@@ -0,0 +1,42 @@
...
+
+x-logging:
+  &default-logging
+  driver: journald
+  options:
+    tag: "umbrel-app {{.Name}}"
+
+services:
+  web:
+    image: yzernik/squeaknode:v0.1.21@sha256:8555988fd28b4d9e6533827510f8f618182d10de3f6e06b988d89048617615ba
```
lukechilds, line 11: The container isn't stopping when you run `scripts/app stop squeaknode`. It just runs for the 1m for the `stop_grace_period` and then gets killed. Probably because signals aren't being passed through, wrapping in a mini init system should resolve it.

```suggestion
    image: yzernik/squeaknode:v0.1.21@sha256:8555988fd28b4d9e6533827510f8f618182d10de3f6e06b988d89048617615ba
    init: true
```

apps/squeaknode/docker-compose.yml
```diff
@@ -0,0 +1,42 @@
+version: "3.7"
+
+x-logging:
+  &default-logging
+  driver: journald
+  options:
+    tag: "umbrel-app {{.Name}}"
+
+services:
+  web:
```
lukechilds, line 10: The container is currently run as root, it should be run as a non-root user.

```suggestion
  web:
    user: "1000:1000"
```

apps/squeaknode/docker-compose.yml
```diff
@@ -0,0 +1,42 @@
...
+    image: yzernik/squeaknode:v0.1.21@sha256:8555988fd28b4d9e6533827510f8f618182d10de3f6e06b988d89048617615ba
+    logging: *default-logging
+    restart: on-failure
+    stop_grace_period: 1m
+    ports:
+      - 8774:8774
```
lukechilds, line 16: I see the web ui is exposed at port 12994. What is port 8774 used for? Does it need to be open to the local network? If not you can remove this line.

apps/squeaknode/docker-compose.yml
```diff
@@ -0,0 +1,42 @@
...
+      - 12994:12994
+    volumes:
+      - ${APP_DATA_DIR}/sqk:/sqk
```
lukechilds, line 19: This directory should be committed into source control, even if it's empty.

Git doesn't actually allow empty directories so you can touch a file at `sqk/.gitkeep` and commit it.

When installing your app Umbrel will strip out the `.gitkeep` file.

The reason to do this is that if the directory doesn't exist Docker will create it the first time your container is run as root and only processes running as root will be able to read from it.

If you commit the dir you can set the correct permissions and they will be respected.

lukechilds (changes requested):
Submitted some feedback on the Umbrel specific integration.

Squeak Node seems like a cool idea! I'm not sure I fully grok exactly how it works yet but I left you some questions about that on Keybase.

apps/squeaknode/docker-compose.yml
lukechilds: Ahh I just read through your docs and realised this port needs to be exposed for remote peers to connect to your. You don't need to expose this in the compose file, that's just to open it up on the local network.

We add a Tor hidden service for each app to allow remote access. Is it possible for Squeak Node to connect to other node's via Tor hidden services?

If not it should be possible by accepting a SOCKS5 proxy as part of your config and then pointing it at the Tor proxy.

stevenroose:
@yzernik just a heads up that if you don't have explicit individual approval by all the contributors to this repository to publish your modified source code ("Adapted Material"), you are officially in violation of the license of this project and you can be sued:

apps/squeaknode/docker-compose.yml
yzernik: ✓ added the empty sqk directory with gitkeep

apps/squeaknode/docker-compose.yml
yzernik: I did some debugging and found the orphan thread that was stopping the clean shutdown. I think I can skip the init option.

apps/squeaknode/docker-compose.yml
yzernik: I would definitely like to use Tor hidden service for my app. That would be perfect. I don't know where to add the SOCKS5 config...

apps/squeaknode/docker-compose.yml
```diff
@@ -0,0 +1,44 @@
...
+services:
+  web:
+    image: yzernik/squeaknode:v0.1.89@sha256:249337c965855df5bd883a680ff3172993e2273415e63671e31ed4df6850ae5a
+    restart: on-failure
+    stop_grace_period: 1m
+    ports:
+      - 8994:8994
+      - 12994:12994
+      - 8555:8555
+      - 18555:18555
```
lukechilds, line 12: Are you able to explain (or link me to some docs) regarding what each port is used for?

apps/squeaknode/docker-compose.yml
```diff
@@ -0,0 +1,44 @@
...
+      SQUEAKNODE_WEBADMIN_ENABLED: "true"
+      SQUEAKNODE_WEBADMIN_USERNAME: "umbrel"
+      SQUEAKNODE_WEBADMIN_PASSWORD: "moneyprintergobrrr"
+
+      SQUEAKNODE_CORE_NETWORK: "$BITCOIN_NETWORK"
+      SQUEAKNODE_CORE_SQK_DIR_PATH: "/sqk"
+
+      DEBUG: "squeaknode:*"
+
+  envoy:
```
lukechilds, line 40: I noticed you aren't exposing any ports from this container, what is it being used for?

apps/squeaknode/docker-compose.yml
```diff
@@ -0,0 +1,44 @@
...
+    image: envoyproxy/envoy:v1.16.0
+    network_mode: "host"
```
lukechilds, line 42: You should expose the explicit ports you need to open instead of giving envoy host networking access here.

yzernik: Fixed.

apps/squeaknode/docker-compose.yml
yzernik: The envoy container is used to bridge the gRPC backend with the gRPC-web frontent.
https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/grpc_web_filter

apps/squeaknode/docker-compose.yml
yzernik: 8994 is used by the gRPC server.
12994 is used by the web server.
8555 is used by the p2p protocol in mainnet.
18555 is used by the p2p protocol in testnet.

yzernik: I removed ports 8555 and 18555, since they will be accessible through the tor configuration.

yzernik:
Web request to envoy is currently failing with this message:
`Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at http://3vswdgskvj7hza5xwfdxlj6yh7lc6n66bnukobejfgfm2qpzfll3wsad.onion:15081/squeaknode.SqueakAdmin/GetTimelineSqueakDisplays. (Reason: CORS header ‘Access-Control-Allow-Origin’ does not match ‘null’).`

lukechilds:
Merged 🎉

Thanks so much for your contribution @yzernik, looking forward to seeing Umbrel users play with Squeaknode!

Issue #109, opened 2020-08-09
Update Umbrel OS systemd services via OTA updates


mayankchhabra:

AaronDewes:
Fixed by #638.

nmfretz:
> Fixed by https://github.com/getumbrel/umbrel/pull/638.

Closing this issue

PR #387, opened 2021-01-01
Run karen as systemd service

AaronDewes:
Fixes https://github.com/orgs/getumbrel/projects/3#card-50139347

This also makes sure that systemd services are always updated if Umbrel is updated.

This works because karen refuses to start if there already is one (two Karens in one store are too much), so the start script doesn't start karen.

### DIFF
--- a/scripts/start
+++ b/scripts/start
@@ -71,7 +71,7 @@ cd "$UMBREL_ROOT"
 
 echo "Starting karen..."
 echo
-./karen &>> "${UMBREL_LOGS}/karen.log" &
+./scripts/karen
 
 echo "Starting backup monitor..."
 echo
--- a/scripts/karen
+++ b/scripts/karen
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+# Simple wrapper script for karen to log to the correct logfile
+
+UMBREL_ROOT="$(dirname $(readlink -f "${BASH_SOURCE[0]}"))/.."
+UMBREL_LOGS="${UMBREL_ROOT}/logs"
+
+./karen &>> "${UMBREL_LOGS}/karen.log" &
--- a/scripts/umbrel-os/services/karen.service
+++ b/scripts/umbrel-os/services/karen.service
@@ -1,0 +1,22 @@
+# Umbrel Karen Service
+# Installed at /etc/systemd/system/karen.service
+
+[Unit]
+Description=Umbrel Karen Service
+Requires=umbrel-external-storage.service
+After=umbrel-external-storage.service
+
+[Service]
+Type=forking
+TimeoutStartSec=infinity
+TimeoutStopSec=16min
+ExecStart=/home/umbrel/umbrel/scripts/karen
+User=root
+Group=root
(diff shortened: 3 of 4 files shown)


scripts/update/01-run.sh
```diff
@@ -27,15 +27,13 @@ if [[ ! -z "${UMBREL_OS:-}" ]]; then
...
-    if [[ $UMBREL_OS == "v0.1.2" ]] && [[ -f "/etc/avahi/avahi-daemon.conf" ]]; then
-        echo "Binding Avahi to eth0 and wlan0"
-        sed -i "s/#allow-interfaces=eth0/allow-interfaces=eth0,wlan0/g;" "/etc/avahi/avahi-daemon.conf"
-        systemctl restart avahi-daemon.service
-    fi
+    # Older versions of Umbrel OS (<0.3.1) didn't have karen as systemd service
+    # This installs the new karen systemd service and makes sure systemd services are always updated.
+    UMBREL_SYSTEMD_SERVICES=$(ls ${UMBREL_ROOT}/.umbrel-${RELEASE}/scripts/umbrel-os/services/*.service)
+    for service in $UMBREL_SYSTEMD_SERVICES; do
+      install -m 644 "${service}"   "/etc/systemd/system/${service}"
```
louneskmt, line 34:
```suggestion
      install -m 644 "${service}" "/etc/systemd/system/${service}"
```

scripts/start
```diff
@@ -67,19 +67,20 @@ export DEVICE_HOSTNAME="${DEVICE_HOSTNAME}.local"
 export DOCKER_CLIENT_TIMEOUT=240
 export COMPOSE_HTTP_TIMEOUT=240
 
-cd "$UMBREL_ROOT"
 
 echo "Starting karen..."
 echo
-./karen &>> "${UMBREL_LOGS}/karen.log" &
+./karen
```
louneskmt, line 73: I'm not really sure why you removed the `cd` command here. You could have just executed karen from the scripts dir:

```suggestion
cd "$UMBREL_ROOT"

echo "Starting karen..."
echo
./scripts/karen
```

Also, karen is already started by the system before the `start` script (as you specified in the `umbrel-startup`), so why do you manually execute it here? We can just check if it is Umbrel OS and start the karen service if it's not started (e.g. if a user manually use this script), and if it is not Umbrel OS, start karen manually. 

What do you think?

AaronDewes: > Also, on Umbrel OS karen is already started by the system before the start script (as you specified in the umbrel-startup), so why do you manually execute it here? 

karen itself already checks if it's running, so it just refuses to start and we dohn't have an additional check required here.

> I'm not really sure why you removed the cd command here. 

Because the two commands after the cd were run fromvthe script dir, only the two commands before were not, but I can revert that if you want.

scripts/update/01-run.sh
AaronDewes: That's the same in Umbrel OS, so that style fix can also be applied there.

AaronDewes:
I'll split this PR, btw, it fixes #109

PR #388, opened 2021-01-01
Fix lncli and bitcoin-cli wrapper scripts

AaronDewes:
I assume the problem was the .env file not being loaded correctly. By temporarily going into the Umbrel root folder and going back later, the problem can be solved.


lukechilds:
Hey Aaron, what was the issue here?

The CLI scripts are working fine for me without this change.

lukechilds (approved):
Ok so as discussed on Telegram the issue occurs when executing the scripts from outside `UMBREL_ROOT`.

Just tweaked the code to pass in the env instead of changing the dir.

Great spot, thanks!

PR #396, opened 2021-01-04
Adding check if drive has enough space

ronnicek:
Adding check if drive has enough space

### DIFF
--- a/scripts/configure
+++ b/scripts/configure
@@ -48,6 +48,14 @@ if [ "$BITCOIN_NETWORK" != "mainnet" ] && [ "$BITCOIN_NETWORK" != "testnet" ] &&
   exit 1
 fi
 
+if [ "$BITCOIN_NETWORK" == "mainnet" ] && [ ! -f "${STATUS_DIR}/configured" ]; then
+  # check size of drive when mainnet is used and Umbrel is not configured
+  # check if drive has more then 700GB
+  if [[ $(df -k "$UMBREL_ROOT" | grep -v "Available" | awk '{ print $4 }') -lt 700000000 ]]; then
+    echo "Not enough space on drive - 1TB drive needed"
+  fi
+fi
+
 # Get current Umbrel version
 UMBREL_VERSION="$(cat ${UMBREL_ROOT}/info.json | jq -r .version)"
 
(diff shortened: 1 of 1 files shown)


scripts/configure
```diff
@@ -48,6 +48,14 @@ if [ "$BITCOIN_NETWORK" != "mainnet" ] && [ "$BITCOIN_NETWORK" != "testnet" ] &&
   exit 1
 fi
 
+if [ "$BITCOIN_NETWORK" == "mainnet" ] && [ ! -f "${STATUS_DIR}/configured" ]; then
+  # check size of drive when mainnet is used and Umbrel is not configured
+  # check if drive has more then 700GB
+  if [[ $(df -k "$UMBREL_ROOT" | grep -v "Available" | awk '{ print $4 }') -lt 700000000 ]]; then
+    echo "Not enough space on drive - 1TB drive needed"
+  fi
```
AaronDewes, line 56: I think the script should write some status file that the manager can read and notify the dashboard. Your current script just does nothing if no disk space is free except print a warning (and many Umbrel OS users won't notice that)

```suggestion
if [ "$BITCOIN_NETWORK" == "mainnet" ] && [ ! -f "${STATUS_DIR}/configured" ]; then
  # check size of drive when mainnet is used and Umbrel is not configured
  # check if drive has more then 700GB
  if [[ $(df -k "$UMBREL_ROOT" | grep -v "Available" | awk '{ print $4 }') -lt 700000000 ]]; then
    echo "Not enough space on drive - 1TB drive needed"
    touch ""$UMBREL_ROOT/statuses/disk-too-small"
  fi
```

@louneskmt @lukechilds @louneskmt  Do you tihnk that's a good way to notify users and can one of you implement that on the dashboard?

AaronDewes (changes requested):

scripts/configure
ronnicek: Maybe.. I created mainly for people which are running it with manual installation. Don't know how exactly UmbrelOS is working - if it's using the same, then I think it would be great idea 👍 

Problem is, that I don't know how to implement it in dashboard :-(

Also question is, if use 700GB or 600GB (since this is in howto) should be the minimum.

AaronDewes: You don't need to change anything for Umbrel OS. For the dashboard, just add `touch "$UMBREL_ROOT/statuses/disk-too-small"``as I suggested above, either@mayankchhabra, @lukechilds or @louneskmt will then be able to implement it on the dashboard. They'll also know what to use as minimum

AaronDewes (approved):

mroxso (approved):

mayankchhabra:
Thanks a lot for working on this @ronnicek! As you might have noticed, quite a lot has changed recently with Umbrel 0.5 and it's no longer required to have a 1TB drive. 😅

Closing this PR now. My apologies for the resulting unproductive hard work. :(

PR #400, opened 2021-01-08
Update specter-desktop app to v1.0.0

nolim1t:
Update Specter Desktop to [v1.0.0](https://github.com/cryptoadvance/specter-desktop/releases/tag/v1.0.0)

/cc @k9ert

### DIFF
--- a/apps/registry.json
+++ b/apps/registry.json
@@ -27,7 +27,7 @@
         "id": "specter-desktop",
         "category": "Wallets",
         "name": "Specter Desktop",
-        "version": "0.10.4",
+        "version": "1.0.0",
         "tagline": "Multisig with hardware wallets made easy",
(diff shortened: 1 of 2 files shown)


k9ert:
So it seems that concerning the ssl-question, more people than i've expected think that there shouldn't be a self-signed-cert. I'm not yet convinced but i commit to that!

> It seems easier and less dangerous to instead tell users how to just navigate to chrome://flags/#unsafely-treat-insecure-origin-as-secure enter http://umbrel.local:25441 in the text box, and then restart Chrome.

I'm not sure it's that easy with firefow, other browsers are not supported anyway.

> We could maybe add the instructions for this to the Umbrel UI. We could also detect the browser and provide different instructions for Firefox etc.

Yes, would be cool to integrate that into umbrel in some way.

This PR is imho ready to be merged!

nolim1t:
> So it seems that concerning the ssl-question, more people than i've expected think that there shouldn't be a self-signed-cert. I'm not yet convinced but i commit to that!
> 
> > It seems easier and less dangerous to instead tell users how to just navigate to chrome://flags/#unsafely-treat-insecure-origin-as-secure enter http://umbrel.local:25441 in the text box, and then restart Chrome.
> 
> I'm not sure it's that easy with firefow, other browsers are not supported anyway.
> 
> > We could maybe add the instructions for this to the Umbrel UI. We could also detect the browser and provide different instructions for Firefox etc.
> 
> Yes, would be cool to integrate that into umbrel in some way.
> 
> This PR is imho ready to be merged!

I think, self signed certificates is not a very good user experience (configuring browser stuff).

nolim1t:
You can review some of the changes then someone can merge it in

k9ert:
> I think, self signed certificates is not a very good user experience (configuring browser stuff).
Definitely agree to that. But getting breached which could potentially been prevented with a self-signe-cert is also not a very good UX. It's clearly a non-trivial trade-off-decision.

lukechilds (approved):
Awesome, thanks for this @nolim1t!

Just FYI if you're submitting any app updates in the future you need to also bump the version in `apps/registry.json`.

PR #402, opened 2021-01-08
Upgrade thunderhub app to v0.12.2

louneskmt:
This PR upgrades thunderhub app to the latest version v0.12.2.

Changelog:
```
refactor: ♻️ boltz resolvers
chore: 🔧 enable tor proxy config
fix: 🐛 boltz swap
```
https://github.com/apotdevin/thunderhub/releases/tag/v0.12.2

### DIFF
--- a/apps/registry.json
+++ b/apps/registry.json
@@ -119,7 +119,7 @@
         "id": "thunderhub",
         "category": "Lightning Node Management",
         "name": "ThunderHub",
-        "version": "0.12.1",
+        "version": "0.12.2",
         "tagline": "Take full control of your Lightning node",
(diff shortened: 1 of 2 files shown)


lukechilds (approved):
Thanks @louneskmt! Tested and working great!

lukechilds:
> chore: 🔧 enable tor proxy config

I wonder if that's something we can make use of in Umbrel?

cc @apotdevin

apotdevin:
Definitely! You just need to pass the env var `TOR_PROXY_SERVER='socks://your.tor.endpoint'` and thunderhub will proxy all the fetch requests it does, like price api and blockchain fees through tor 😎

[Here](https://github.com/apotdevin/thunderhub#tor-requests) is the doc for this.

Issue #403, opened 2021-01-09
Installing New App Deleted all other Apps (and took me to "what is your name" initial setup screen)


maltokyo:
Something went very wrong today.  I am on Umbrel V0.3.1.

I pressed install on Sphinx, and it deleted all other install apps (BTCPay and Specter).  There was simply nothing installed in the apps section after this.

And now I have this screen when I log in to Umbrel (after I tried to shut it down and reboot).  Have I lost everything?

![image](https://user-images.githubusercontent.com/1388507/104086423-34b7ea00-5258-11eb-937f-74a0aee46cb1.png)

maltokyo:
Debug log - the "jq errors" look to be the cause:

```
# ./scripts/debug
=====================
= Umbrel debug info =
=====================

Umbrel version
--------------
0.3.1

Filesystem information
----------------------
Filesystem      Size  Used Avail Use% Mounted on
udev            1.9G     0  1.9G   0% /dev
tmpfs           384M   16M  368M   5% /run
/dev/sda2       934G  458G  429G  52% /
tmpfs           1.9G     0  1.9G   0% /dev/shm
tmpfs           5.0M     0  5.0M   0% /run/lock
tmpfs           1.9G     0  1.9G   0% /sys/fs/cgroup
/dev/sda1       511M  5.2M  506M   1% /boot/efi
overlay         934G  458G  429G  52% /var/lib/docker/overlay2/850b3a5febafa2c758ef431d64a9bb0a267e143698bd9311cdf9e548fe0946c2/merged
tmpfs           384M     0  384M   0% /run/user/0
overlay         934G  458G  429G  52% /var/lib/docker/overlay2/02570a889b9bb8575c9a3d271104e6b26b0523dd4a795e72c2ad9967c400102f/merged
overlay         934G  458G  429G  52% /var/lib/docker/overlay2/f794d84d2bc4e876e3e5a502adb12a847f97810178af9fa08a843c0dee514de7/merged
overlay         934G  458G  429G  52% /var/lib/docker/overlay2/a1b44ac58d1ccf40485d289a89cdcb0bafca293c4fb566b5245d2a47a5ebd0c6/merged
overlay         934G  458G  429G  52% /var/lib/docker/overlay2/6045c5d569ce0d028f4764db6de5fe8a3a178efec748e7cb7df632c78c572986/merged
[...]

PR #405, opened 2021-01-09
Ensure thread safety when updating app entries

AaronDewes:
See #403

### DIFF
--- a/scripts/app
+++ b/scripts/app
@@ -99,11 +99,11 @@ compose() {
 }
 
 update_installed_apps() {
-    if [[ -f "${USER_FILE}.lock" ]]; then
-      echo "The user file is locked, waiting for it to unlock..."
-      while [[ -f "${USER_FILE}.lock" ]]; do
-          sleep 1
-      done
+  if [[ -f "${USER_FILE}.lock" ]]; then
+    echo "The user file is locked, waiting for it to unlock..."
+    while [[ -f "${USER_FILE}.lock" ]]; do
+      sleep 1
+    done
   fi
   touch "${USER_FILE}.lock"
   local action="${1}"
--- a/scripts/app
+++ b/scripts/app
@@ -99,9 +99,9 @@ compose() {
 }
 
 update_installed_apps() {
-    if [[ ! -f "${USER_FILE}.lock" ]]; then
+    if [[ -f "${USER_FILE}.lock" ]]; then
       echo "The user file is locked, waiting for it to unlock..."
-      while [[ ! -f "${USER_FILE}.lock" ]]; do
+      while [[ -f "${USER_FILE}.lock" ]]; do
           sleep 1
       done
   fi
--- a/scripts/app
+++ b/scripts/app
@@ -4,7 +4,6 @@ set -euo pipefail
 UMBREL_ROOT="$(readlink -f $(dirname "${BASH_SOURCE[0]}")/..)"
 USER_FILE="${UMBREL_ROOT}/db/user.json"
 
-
 show_help() {
   cat << EOF
 app 0.0.1
@@ -22,7 +21,6 @@ Commands:
     ls-installed               Lists installed apps
 EOF
 }
-
 
 check_if_not_already_running() {
   if ps ax | grep $0 | grep -v $$ | grep bash | grep -v grep
@@ -117,9 +115,12 @@ update_installed_apps() {
 }
 
 # Make sure no other install script is running, if it is, wait for it to exit
-while [[ "$(check_if_not_already_running)" == "2" ]]; do 
-    sleep 1
-done
+if [[ "$(check_if_not_already_running)" == "2" ]]; then
+    echo "Another instance of the app script is already running, waiting for it to finish..."
+    while [[ "$(check_if_not_already_running)" == "2" ]]; do
+        sleep 1
+    done
+fi
 
 # Pulls down images for an app and starts it
 if [[ "$command" = "install" ]]; then
--- a/scripts/app
+++ b/scripts/app
@@ -117,7 +117,7 @@ update_installed_apps() {
 }
 
 # Make sure no other install script is running, if it is, wait for it to exit
-while [ $(check_if_not_already_running) == 2]; do 
+while [[ "$(check_if_not_already_running)" == "2" ]]; do 
     sleep 1
 done
 
--- a/scripts/app
+++ b/scripts/app
@@ -124,14 +124,6 @@ update_installed_apps() {
   fi
 }
 
-update_installed_apps_sub() {
-  local action="${1}"
-  local app="${2}"
-  [[ "${action}" == "add" ]] && operator="+" || operator="-"
-  updated_json=$(cat "${USER_FILE}" | jq ".installedApps |= (. ${operator} [\"${app}\"] | unique)")
-  echo "${updated_json}" > "${USER_FILE}"
-}
-
 # Pulls down images for an app and starts it
 if [[ "$command" = "install" ]]; then
 
--- a/scripts/app
+++ b/scripts/app
@@ -22,14 +22,6 @@ Commands:
 EOF
 }
 
-check_if_not_already_running() {
-  if ps ax | grep $0 | grep -v $$ | grep bash | grep -v grep
-  then
-    return 2
-  fi
-  return 0
-}
-
 check_dependencies () {
   for cmd in "$@"; do
     if ! command -v $cmd >/dev/null 2>&1; then
@@ -107,20 +99,20 @@ compose() {
 }
 
 update_installed_apps() {
+    if [[ ! -f "${USER_FILE}.lock" ]]; then
+      echo "The user file is locked, waiting for it to unlock..."
+      while [[ ! -f "${USER_FILE}.lock" ]]; do
+          sleep 1
+      done
+  fi
+  touch "${USER_FILE}.lock"
   local action="${1}"
   local app="${2}"
   [[ "${action}" == "add" ]] && operator="+" || operator="-"
   updated_json=$(cat "${USER_FILE}" | jq ".installedApps |= (. ${operator} [\"${app}\"] | unique)")
   echo "${updated_json}" > "${USER_FILE}"
+  touch "${USER_FILE}.lock"
 }
-
-# Make sure no other install script is running, if it is, wait for it to exit
-if [[ "$(check_if_not_already_running)" == "2" ]]; then
-    echo "Another instance of the app script is already running, waiting for it to finish..."
-    while [[ "$(check_if_not_already_running)" == "2" ]]; do
-        sleep 1
-    done
-fi
 
 # Pulls down images for an app and starts it
 if [[ "$command" = "install" ]]; then
--- a/scripts/app
+++ b/scripts/app
@@ -4,6 +4,7 @@ set -euo pipefail
 UMBREL_ROOT="$(readlink -f $(dirname "${BASH_SOURCE[0]}")/..)"
 USER_FILE="${UMBREL_ROOT}/db/user.json"
 
+
 show_help() {
   cat << EOF
 app 0.0.1
@@ -20,6 +21,15 @@ Commands:
     compose                    Passes all arguments to docker-compose
     ls-installed               Lists installed apps
 EOF
+}
+
+
+check_if_not_already_running() {
+  if ps ax | grep $0 | grep -v $$ | grep bash | grep -v grep
+  then
+    return 2
+  fi
+  return 0
 }
 
 check_dependencies () {
@@ -105,6 +115,11 @@ update_installed_apps() {
   updated_json=$(cat "${USER_FILE}" | jq ".installedApps |= (. ${operator} [\"${app}\"] | unique)")
   echo "${updated_json}" > "${USER_FILE}"
 }
+
+# Make sure no other install script is running, if it is, wait for it to exit
+while [ $(check_if_not_already_running) == 2]; do 
+    sleep 1
+done
 
 # Pulls down images for an app and starts it
 if [[ "$command" = "install" ]]; then
--- a/scripts/app
+++ b/scripts/app
@@ -99,17 +99,17 @@ compose() {
 }
 
 update_installed_apps() {
+  local action="${1}"
+  local app="${2}"
+
   while ! (set -o noclobber; echo "$$" > "${USER_FILE}.lock") 2> /dev/null; do
-    echo "File locked!"
-    echo "Waiting for it to be unlocked..."
+    echo "Waiting for JSON lock to be released for ${app} update..."
     sleep 1
   done
   # This will cause the lock-file to be deleted in case of a
   # premature exit.
   trap "rm -f "${USER_FILE}.lock"; exit $?" INT TERM EXIT
 
-  local action="${1}"
-  local app="${2}"
   [[ "${action}" == "add" ]] && operator="+" || operator="-"
   updated_json=$(cat "${USER_FILE}" | jq ".installedApps |= (. ${operator} [\"${app}\"] | unique)")
   echo "${updated_json}" > "${USER_FILE}"
--- a/scripts/app
+++ b/scripts/app
@@ -100,21 +100,28 @@ compose() {
 
 update_installed_apps() {
   local lockfile="${USER_FILE}.lock"
-  if (set -o noclobber; echo "$$" > "$lockfile") 2> /dev/null;  then
+  if (set -o noclobber; echo "$$" > "$lockfile") 2> /dev/null; then
+    echo "File locked!" 
+    echo "Waiting for it to be unlocked..."
+    sleep 1
     # This will cause the lock-file to be deleted in case of a
     # premature exit.
     trap 'rm -f "$lockfile"; exit $?' INT TERM EXIT
 
-    update_installed_apps_sub $@
+    local action="${1}"
+    local app="${2}"
+    [[ "${action}" == "add" ]] && operator="+" || operator="-"
+    updated_json=$(cat "${USER_FILE}" | jq ".installedApps |= (. ${operator} [\"${app}\"] | unique)")
+    echo "${updated_json}" > "${USER_FILE}"
 
     rm -f "$lockfile"
     trap - INT TERM EXIT
-else
-   echo "File locked!" 
-   echo "Waiting for it to be unlocked..."
-   sleep 1
-   update_installed_apps $@
-fi
+  else
+    echo "File locked!" 
+    echo "Waiting for it to be unlocked..."
+    sleep 1
+    update_installed_apps $@
+  fi
 }
 
 update_installed_apps_sub() {
--- a/scripts/app
+++ b/scripts/app
@@ -99,9 +99,8 @@ compose() {
 }
 
 update_installed_apps() {
-  until (set -o noclobber; echo "$$" > "${USER_FILE}.lock") 2> /dev/null
-  do
-    echo "File locked!" 
+  while ! (set -o noclobber; echo "$$" > "${USER_FILE}.lock") 2> /dev/null; do
+    echo "File locked!"
     echo "Waiting for it to be unlocked..."
     sleep 1
   done
--- a/scripts/app
+++ b/scripts/app
@@ -99,29 +99,23 @@ compose() {
 }
 
 update_installed_apps() {
-  local lockfile="${USER_FILE}.lock"
-  if (set -o noclobber; echo "$$" > "$lockfile") 2> /dev/null; then
+  until (set -o noclobber; echo "$$" > "${USER_FILE}.lock") 2> /dev/null
+  do
     echo "File locked!" 
     echo "Waiting for it to be unlocked..."
     sleep 1
-    # This will cause the lock-file to be deleted in case of a
-    # premature exit.
-    trap 'rm -f "$lockfile"; exit $?' INT TERM EXIT
+  done
+  # This will cause the lock-file to be deleted in case of a
+  # premature exit.
+  trap "rm -f "${USER_FILE}.lock"; exit $?" INT TERM EXIT
 
-    local action="${1}"
-    local app="${2}"
-    [[ "${action}" == "add" ]] && operator="+" || operator="-"
-    updated_json=$(cat "${USER_FILE}" | jq ".installedApps |= (. ${operator} [\"${app}\"] | unique)")
-    echo "${updated_json}" > "${USER_FILE}"
+  local action="${1}"
+  local app="${2}"
+  [[ "${action}" == "add" ]] && operator="+" || operator="-"
+  updated_json=$(cat "${USER_FILE}" | jq ".installedApps |= (. ${operator} [\"${app}\"] | unique)")
+  echo "${updated_json}" > "${USER_FILE}"
 
-    rm -f "$lockfile"
-    trap - INT TERM EXIT
-  else
-    echo "File locked!" 
-    echo "Waiting for it to be unlocked..."
-    sleep 1
-    update_installed_apps $@
-  fi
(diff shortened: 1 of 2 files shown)


scripts/app
```diff
@@ -106,6 +116,11 @@ update_installed_apps() {
   echo "${updated_json}" > "${USER_FILE}"
 }
 
+# Make sure no other install script is running, if it is, wait for it to exit
+while [ $(check_if_not_already_running) == 2]; do
```
lukechilds, line 120:
```suggestion
while [[ "$(check_if_not_already_running)" == "2" ]]; do 
```

Some suggestions:
- Prefer `[[` over `[` to prevent weird behaviour
- Always quote strings in bash to prevent weird behaviour

scripts/app
```diff
@@ -106,6 +116,11 @@ update_installed_apps() {
   echo "${updated_json}" > "${USER_FILE}"
 }
 
+# Make sure no other install script is running, if it is, wait for it to exit
+while [ $(check_if_not_already_running) == 2]; do 
+    sleep 1
```
lukechilds, line 121: It would be nice to print something here like:

```
app script is already running, sleeping for 1 second...
```

Or maybe it's neater to do the log before the loop like:

```bash
if [[ "$(check_if_not_already_running)" == "2" ]]; do 
  echo "Another instance of the app script is already running, waiting for it to finish..."
  # Now do wait loop
  echo "Resuming!"
fi
```

so the logs don't get spammed.

lukechilds (changes requested):
Thanks so much for looking in to this Aaron!

Some tiny suggestions.

AaronDewes:
Done. Additionally, it would be possible to update the version of the script (https://github.com/getumbrel/umbrel/pull/405/files#diff-8e674b8e14b37a67c77e07bc9ad10ace548b82c5729fe2c679001b6eedd4dd41R9)

maltokyo:
I still frequently get plagued by this issue, any chances it could be merged soon.  @lukechilds 
I know you're super busy, and locking down security or other things should be a priority, but some bugs like this one seem to wait a long time to get merged.

lukechilds:
@AaronDewes are you able to reproduce this reliably?

Are you able to provide me with a script I can run to reliably trigger the bug?

maltokyo:
@lukechilds for me it happens when I have one app installed.  Then install another one right after the first one.  But sometimes it does not happen, and I don't know what the difference is.

AaronDewes:
> @AaronDewes are you able to reproduce this reliably?
> 
> Are you able to provide me with a script I can run to reliably trigger the bug?
I haven't found a way to do this, as the exact speed of the app removal isn't an exact value, you could try creating a bash script that starts the removal of the same app twice, like

```
sudo ./scripts/app uninstall app &
sudo ./scripts/app uninstall app &
```
But there is no way to be sure this bug is triggered, there is a really small timeframe that this can happen.

scripts/app
```diff
@@ -22,6 +22,14 @@ Commands:
 EOF
 }
 
+check_if_not_already_running() {
+  if ps ax | grep $0 | grep -v $$ | grep bash | grep -v grep
```
lukechilds, line 26: Is this a reliable check?

Won't it match any process that contains the string `app`?

Maybe it would be better to use a lockfile.

lukechilds: Also would be good to stick to the current convention of `if condition; then` all on the same line.

scripts/app
```diff
@@ -106,6 +114,14 @@ update_installed_apps() {
   echo "${updated_json}" > "${USER_FILE}"
 }
 
+# Make sure no other install script is running, if it is, wait for it to exit
+if [[ "$(check_if_not_already_running)" == "2" ]]; then
```
lukechilds, line 118: Rather than halting the execution of the entire script, which prevents apps from being installed in parallel, we could maybe check if the JSON file is currently being updated with a write lock, which allows apps to be installed/uninstalled in parallel and just briefly blocked while waiting to do the write.

lukechilds (changes requested):

scripts/app
AaronDewes: > Also would be good to stick to the current convention of if condition; then all on the same line.

I copied this check from karen without changes, you should probably fix it there

scripts/app
```diff
@@ -99,11 +99,19 @@ compose() {
...
+          sleep 1
+      done
+  fi
+  touch "${USER_FILE}.lock"
   local action="${1}"
   local app="${2}"
   [[ "${action}" == "add" ]] && operator="+" || operator="-"
   updated_json=$(cat "${USER_FILE}" | jq ".installedApps |= (. ${operator} [\"${app}\"] | unique)")
   echo "${updated_json}" > "${USER_FILE}"
+  touch "${USER_FILE}.lock"
```
louneskmt, line 114: This should remove the lock file instead, shouldn't it?

```suggestion
  rm "${USER_FILE}.lock"
```

maltokyo:
i cannot reproduce this reliably @lukechilds @AaronDewes but it happened again today on my Umbrel.  No idea what is the timing that can reproduce it (or what can reliably prevent it), I hope this fix works

lukechilds:
@AaronDewes 

>I haven't found a way to do this, as the exact speed of the app removal isn't an exact value, you could try creating a bash script that starts the removal of the same app twice, like
>
>```
>sudo ./scripts/app uninstall app &
>sudo ./scripts/app uninstall app &
>```
>But there is no way to be sure this bug is triggered, there is a really small timeframe that this can happen.

So the bug is only hit when the same app is uninstalled twice at the same time?

I'm trying really hard to re-create this but I'm unable to. I want to make sure we understand the bug and what's causing it before we merge the fix. In my testing I'm running `install`/`uninstall` loads of times async in a loop and I never end up with invalid JSON. Sometimes apps are dropped if the writes happen async (which was always expected but not really an issue for beta) but I never end up with invalid JSON.

I'm using this user.json:

```json
{
  "name": "satoshi",
  "password": "password",
  "seed": "seed",
  "installedApps": [
    "foo"
  ]
}
```

And this modified `app` script that just reads the JSON file and writes the app value back to it, nothing else:

```bash
#!/usr/bin/env bash
set -euo pipefail

UMBREL_ROOT="$(readlink -f $(dirname "${BASH_SOURCE[0]}"))"
USER_FILE="${UMBREL_ROOT}/user.json"

command="$1"
app="$2"

update_installed_apps() {
  local action="${1}"
  local app="${2}"
  [[ "${action}" == "add" ]] && operator="+" || operator="-"
  updated_json=$(cat "${USER_FILE}" | jq ".installedApps |= (. ${operator} [\"${app}\"] | unique)")
  echo "${updated_json}" > "${USER_FILE}"
}

if [[ "$command" = "install" ]]; then
  update_installed_apps add "${app}"
  exit
fi

if [[ "$command" = "uninstall" ]]; then
  update_installed_apps remove "${app}"
  exit
fi
```

And running `app install` or `app uninstall` or even both install and uninstall concurrently never causes a JSON error:

```shell
vagrant@umbrel-dev:/tmp/test$ for i in {1..40}; do ./app install $i & done
vagrant@umbrel-dev:/tmp/test$ cat user.json
{
  "name": "satoshi",
  "password": "password",
  "seed": "seed",
  "installedApps": [
    "1",
    "18",
    "32",
    "6",
    "foo"
  ]
}
vagrant@umbrel-dev:/tmp/test$ for i in {1..40}; do ./app uninstall $i & done
vagrant@umbrel-dev:/tmp/test$ cat user.json
{
  "name": "satoshi",
  "password": "password",
  "seed": "seed",
  "installedApps": [
    "1",
    "18",
    "32",
    "6",
    "foo"
  ]
}
vagrant@umbrel-dev:/tmp/test$ for i in {1..40}; do ./app install app & done
vagrant@umbrel-dev:/tmp/test$ cat user.json
{
  "name": "satoshi",
  "password": "password",
  "seed": "seed",
  "installedApps": [
    "1",
    "18",
    "32",
    "6",
    "app",
    "foo"
  ]
}
vagrant@umbrel-dev:/tmp/test$ for i in {1..40}; do ./app uninstall app & done
vagrant@umbrel-dev:/tmp/test$ cat user.json
{
  "name": "satoshi",
  "password": "password",
  "seed": "seed",
  "installedApps": [
    "1",
    "18",
    "32",
    "6",
    "foo"
  ]
}
vagrant@umbrel-dev:/tmp/test$ for i in {1..40}; do ./app install app & ./app uninstall app &done
vagrant@umbrel-dev:/tmp/test$ cat user.json
{
  "name": "satoshi",
  "password": "password",
  "seed": "seed",
  "installedApps": [
    "1",
    "18",
    "32",
    "6",
    "foo"
  ]
}
vagrant@umbrel-dev:/tmp/test$ for i in {1..40}; do ./app install $i & ./app uninstall $i &done
vagrant@umbrel-dev:/tmp/test$ cat user.json
{
  "name": "satoshi",
  "password": "password",
  "seed": "seed",
  "installedApps": [
    "1",
    "18",
    "22",
    "32",
    "33",
    "35",
    "38",
    "6",
    "foo"
  ]
}
```

Have you been able to recreate this issue? Even if unreliably? Or you haven't encountered it at all and this fix is based on your assumption of what's going wrong?

scripts/app
```diff
@@ -99,11 +99,19 @@ compose() {
 }
 
 update_installed_apps() {
+    if [[ ! -f "${USER_FILE}.lock" ]]; then
```
lukechilds, line 102: Are you able to fix indentation here?

lukechilds, line 102: This is checking if the lock file does not exist, you want to check if it does exist.
```suggestion
    if [[ -f "${USER_FILE}.lock" ]]; then
```

scripts/app
```diff
@@ -99,11 +99,19 @@ compose() {
...
+      echo "The user file is locked, waiting for it to unlock..."
+      while [[ ! -f "${USER_FILE}.lock" ]]; do
```
lukechilds, line 104: Same as above
```suggestion
      while [[ -f "${USER_FILE}.lock" ]]; do
```

lukechilds (changes requested):
These check condition needs inverting otherwise it will hang and never allow an app to be installed.

AaronDewes:
> 
> 
> @AaronDewes
> 
> > I haven't found a way to do this, as the exact speed of the app removal isn't an exact value, you could try creating a bash script that starts the removal of the same app twice, like
> > ```
> > sudo ./scripts/app uninstall app &
> > sudo ./scripts/app uninstall app &
> > ```
> > 
> > 
> > But there is no way to be sure this bug is triggered, there is a really small timeframe that this can happen.
> 
> So the bug is only hit when the same app is uninstalled twice at the same time?
> 
> I'm trying really hard to re-create this but I'm unable to. I want to make sure we understand the bug and what's causing it before we merge the fix. In my testing I'm running `install`/`uninstall` loads of times async in a loop and I never end up with invalid JSON. Sometimes apps are dropped if the writes happen async (which was always expected but not really an issue for beta) but I never end up with invalid JSON.
> 
> I'm using this user.json:
> 
> ```json
> {
>   "name": "satoshi",
>   "password": "password",
>   "seed": "seed",
>   "installedApps": [
>     "foo"
>   ]
> }
> ```
> 
> And this modified `app` script that just reads the JSON file and writes the app value back to it, nothing else:
> 
> ```shell
> #!/usr/bin/env bash
> set -euo pipefail
> 
> UMBREL_ROOT="$(readlink -f $(dirname "${BASH_SOURCE[0]}"))"
> USER_FILE="${UMBREL_ROOT}/user.json"
> 
> command="$1"
> app="$2"
> 
> update_installed_apps() {
>   local action="${1}"
>   local app="${2}"
>   [[ "${action}" == "add" ]] && operator="+" || operator="-"
>   updated_json=$(cat "${USER_FILE}" | jq ".installedApps |= (. ${operator} [\"${app}\"] | unique)")
>   echo "${updated_json}" > "${USER_FILE}"
> }
> 
> if [[ "$command" = "install" ]]; then
>   update_installed_apps add "${app}"
>   exit
> fi
> 
> if [[ "$command" = "uninstall" ]]; then
>   update_installed_apps remove "${app}"
>   exit
> fi
> ```
> 
> And running `app install` or `app uninstall` or even both install and uninstall concurrently never causes a JSON error:
> 
> ```shell
> vagrant@umbrel-dev:/tmp/test$ for i in {1..40}; do ./app install $i & done
> vagrant@umbrel-dev:/tmp/test$ cat user.json
> {
>   "name": "satoshi",
>   "password": "password",
>   "seed": "seed",
>   "installedApps": [
>     "1",
>     "18",
>     "32",
>     "6",
>     "foo"
>   ]
> }
> vagrant@umbrel-dev:/tmp/test$ for i in {1..40}; do ./app uninstall $i & done
> vagrant@umbrel-dev:/tmp/test$ cat user.json
> {
>   "name": "satoshi",
>   "password": "password",
>   "seed": "seed",
>   "installedApps": [
>     "1",
>     "18",
>     "32",
>     "6",
>     "foo"
>   ]
> }
> vagrant@umbrel-dev:/tmp/test$ for i in {1..40}; do ./app install app & done
> vagrant@umbrel-dev:/tmp/test$ cat user.json
> {
>   "name": "satoshi",
>   "password": "password",
>   "seed": "seed",
>   "installedApps": [
>     "1",
>     "18",
>     "32",
>     "6",
>     "app",
>     "foo"
>   ]
> }
> vagrant@umbrel-dev:/tmp/test$ for i in {1..40}; do ./app uninstall app & done
> vagrant@umbrel-dev:/tmp/test$ cat user.json
> {
>   "name": "satoshi",
>   "password": "password",
>   "seed": "seed",
>   "installedApps": [
>     "1",
>     "18",
>     "32",
>     "6",
>     "foo"
>   ]
> }
> vagrant@umbrel-dev:/tmp/test$ for i in {1..40}; do ./app install app & ./app uninstall app &done
> vagrant@umbrel-dev:/tmp/test$ cat user.json
> {
>   "name": "satoshi",
>   "password": "password",
>   "seed": "seed",
>   "installedApps": [
>     "1",
>     "18",
>     "32",
>     "6",
>     "foo"
>   ]
> }
> vagrant@umbrel-dev:/tmp/test$ for i in {1..40}; do ./app install $i & ./app uninstall $i &done
> vagrant@umbrel-dev:/tmp/test$ cat user.json
> {
>   "name": "satoshi",
>   "password": "password",
>   "seed": "seed",
>   "installedApps": [
>     "1",
>     "18",
>     "22",
>     "32",
>     "33",
>     "35",
>     "38",
>     "6",
>     "foo"
>   ]
> }
> ```
> 
> Have you been able to recreate this issue? Even if unreliably? Or you haven't encountered it at all and this fix is based on your assumption of what's going wrong?

I luckily (or sadly for this PR) haven't encountered it myself, I know that it often happned to @maltokyo when installing many apps at the same time, and a few other users. I know for a fact that in every case I've heard of this issue occuring, user.json got currupted with either a duplicated or missing character at the beginning or the end. Also, I know that it only happened when installing multiple apps at the same time, so this is very likely to be this issue.

lukechilds:
Even after inverting the condition, it doesn't seem like this fixes the underlying problem. If multiple installs happen at the same time, they both see the lock file as not existing so both create a lock file at the same time and overwrite eachother.\

using:

```bash
#!/usr/bin/env bash
set -euo pipefail

UMBREL_ROOT="$(readlink -f $(dirname "${BASH_SOURCE[0]}"))"
USER_FILE="${UMBREL_ROOT}/user.json"

command="$1"
app="$2"

update_installed_apps() {
  if [[ -f "${USER_FILE}.lock" ]]; then
    echo "The user file is locked, waiting for it to unlock..."
    while [[ -f "${USER_FILE}.lock" ]]; do
      sleep 1
    done
  fi
  echo creating lockfile
  touch "${USER_FILE}.lock"
  local action="${1}"
  local app="${2}"
  [[ "${action}" == "add" ]] && operator="+" || operator="-"
  updated_json=$(cat "${USER_FILE}" | jq ".installedApps |= (. ${operator} [\"${app}\"] | unique)")
  echo "${updated_json}" > "${USER_FILE}"
  rm "${USER_FILE}.lock"
}

if [[ "$command" = "install" ]]; then
  update_installed_apps add "${app}"
  exit
fi

if [[ "$command" = "uninstall" ]]; then
  update_installed_apps remove "${app}"
  exit
fi
```

```
vagrant@umbrel-dev:/tmp/test$ cat user.json
{
  "name": "satoshi",
  "password": "password",
  "seed": "seed",
  "installedApps": [
    "foo"
  ]
}
vagrant@umbrel-dev:/tmp/test$ for i in {1..10}; do ./app install $i & done
[1] 32606
[2] 32607
[3] 32608
[4] 32609
[5] 32610
[6] 32611
[7] 32612
[8] 32613
[9] 32614
[10] 32615
vagrant@umbrel-dev:/tmp/test$ creating lockfile
creating lockfile
creating lockfile
creating lockfile
The user file is locked, waiting for it to unlock...
The user file is locked, waiting for it to unlock...
The user file is locked, waiting for it to unlock...
The user file is locked, waiting for it to unlock...
The user file is locked, waiting for it to unlock...
The user file is locked, waiting for it to unlock...
rm: rm: cannot remove '/tmp/test/user.json.lock': No such file or directory
cannot remove '/tmp/test/user.json.lock': No such file or directory
rm: cannot remove '/tmp/test/user.json.lock': No such file or directory
creating lockfile
creating lockfile
creating lockfile
creating lockfile
creating lockfile
creating lockfile
rm: cannot remove '/tmp/test/user.json.lock': No such file or directory
rm: cannot remove '/tmp/test/user.json.lock': No such file or directory
rm: cannot remove '/tmp/test/user.json.lock': No such file or directory
rm: cannot remove '/tmp/test/user.json.lock': No such file or directory

[1]   Exit 1                  ./app install $i
[2]   Exit 1                  ./app install $i
[3]   Exit 1                  ./app install $i
[4]   Exit 1                  ./app install $i
[5]   Done                    ./app install $i
[6]   Exit 1                  ./app install $i
[7]   Exit 1                  ./app install $i
[8]   Done                    ./app install $i
[9]-  Done                    ./app install $i
[10]+  Exit 1                  ./app install $i
vagrant@umbrel-dev:/tmp/test$ cat user.json
{
  "name": "satoshi",
  "password": "password",
  "seed": "seed",
  "installedApps": [
    "3",
    "4",
    "5",
    "6",
    "7",
    "8",
    "foo"
  ]
}
```

You can see doing 10 installs in parallel here only 6 get written.

AaronDewes:
> 
> 
> Even after inverting the condition, it doesn't seem like this fixes the underlying problem. If multiple installs happen at the same time, they both see the lock file as not existing so both create a lock file at the same time and overwrite eachother.\
> 
> using:
> 
> ```shell
> #!/usr/bin/env bash
> set -euo pipefail
> 
> UMBREL_ROOT="$(readlink -f $(dirname "${BASH_SOURCE[0]}"))"
> USER_FILE="${UMBREL_ROOT}/user.json"
> 
> command="$1"
> app="$2"
> 
> update_installed_apps() {
>   if [[ -f "${USER_FILE}.lock" ]]; then
>     echo "The user file is locked, waiting for it to unlock..."
>     while [[ -f "${USER_FILE}.lock" ]]; do
>       sleep 1
>     done
>   fi
>   echo creating lockfile
>   touch "${USER_FILE}.lock"
>   local action="${1}"
>   local app="${2}"
>   [[ "${action}" == "add" ]] && operator="+" || operator="-"
>   updated_json=$(cat "${USER_FILE}" | jq ".installedApps |= (. ${operator} [\"${app}\"] | unique)")
>   echo "${updated_json}" > "${USER_FILE}"
>   rm "${USER_FILE}.lock"
> }
> 
> if [[ "$command" = "install" ]]; then
>   update_installed_apps add "${app}"
>   exit
> fi
> 
> if [[ "$command" = "uninstall" ]]; then
>   update_installed_apps remove "${app}"
>   exit
> fi
> ```
> 
> ```
> vagrant@umbrel-dev:/tmp/test$ cat user.json
> {
>   "name": "satoshi",
>   "password": "password",
>   "seed": "seed",
>   "installedApps": [
>     "foo"
>   ]
> }
> vagrant@umbrel-dev:/tmp/test$ for i in {1..10}; do ./app install $i & done
> [1] 32606
> [2] 32607
> [3] 32608
> [4] 32609
> [5] 32610
> [6] 32611
> [7] 32612
> [8] 32613
> [9] 32614
> [10] 32615
> vagrant@umbrel-dev:/tmp/test$ creating lockfile
> creating lockfile
> creating lockfile
> creating lockfile
> The user file is locked, waiting for it to unlock...
> The user file is locked, waiting for it to unlock...
> The user file is locked, waiting for it to unlock...
> The user file is locked, waiting for it to unlock...
> The user file is locked, waiting for it to unlock...
> The user file is locked, waiting for it to unlock...
> rm: rm: cannot remove '/tmp/test/user.json.lock': No such file or directory
> cannot remove '/tmp/test/user.json.lock': No such file or directory
> rm: cannot remove '/tmp/test/user.json.lock': No such file or directory
> creating lockfile
> creating lockfile
> creating lockfile
> creating lockfile
> creating lockfile
> creating lockfile
> rm: cannot remove '/tmp/test/user.json.lock': No such file or directory
> rm: cannot remove '/tmp/test/user.json.lock': No such file or directory
> rm: cannot remove '/tmp/test/user.json.lock': No such file or directory
> rm: cannot remove '/tmp/test/user.json.lock': No such file or directory
> 
> [1]   Exit 1                  ./app install $i
> [2]   Exit 1                  ./app install $i
> [3]   Exit 1                  ./app install $i
> [4]   Exit 1                  ./app install $i
> [5]   Done                    ./app install $i
> [6]   Exit 1                  ./app install $i
> [7]   Exit 1                  ./app install $i
> [8]   Done                    ./app install $i
> [9]-  Done                    ./app install $i
> [10]+  Exit 1                  ./app install $i
> vagrant@umbrel-dev:/tmp/test$ cat user.json
> {
>   "name": "satoshi",
>   "password": "password",
>   "seed": "seed",
>   "installedApps": [
>     "3",
>     "4",
>     "5",
>     "6",
>     "7",
>     "8",
>     "foo"
>   ]
> }
> ```
> 
> You can see doing 10 installs in parallel here only 6 get written.

I'll try another approach that could work better later.

AaronDewes:
It should work now, can you try it?

AaronDewes:
It worked properly for me now

lukechilds:
Dayuuum, nice idea.

I was gonna suggest writing the PID to the lockfile, sleeping another second, then checking the value is still the same PID before continuing. But the noclobber trick is a pretty cool solution.

Yeah, this is preventing dropped writes for me.

Could we refactor to use a while loop instead of recursion calling a sub function?

lukechilds:
Btw how were you testing? With my provided scripts or on a real Umbrel install?

AaronDewes:
> Btw how were you testing? With my provided scripts or on a real Umbrel install?

With your testing script.

> Could we refactor to use a while loop instead of recursion calling a sub function?

Yes, I'll do that.

AaronDewes:
The idea isn't from me, but from a certain website
Stackoverflow

scripts/app
```diff
@@ -99,6 +99,32 @@ compose() {
...
+    trap - INT TERM EXIT
+  else
+    echo "File locked!" 
+    echo "Waiting for it to be unlocked..."
+    sleep 1
+    update_installed_apps $@
+  fi
+}
+
+update_installed_apps_sub() {
```
louneskmt, line 127: You have forgotten to remove the old function.

lukechilds:
@AaronDewes just fyi I meant refactor into a while loop as well as not using recursion.

lukechilds (approved):

lukechilds:
Great work on this @AaronDewes, thanks for looking into it!

I'm still unable to reproduce the bug but if it is being caused by duplicate writes then this should fix it.

Issue #381, opened 2020-12-30
Serious security concern: warning from README is NOT on the website


Kixunil:
The warning about security is not present on getumbrel website. As a result multiple people installed Umbrel without knowing about the security issues. Especially open SSH with default password which I consider a grave issue.

It is good that you have the warning in README but it must be on the website because most people install from the website directly, not from github.

Also perhaps you should remind the users of the issues in your marketing campaigns as they make Umbrel look like finished product.

mayankchhabra:
Thanks for raising your concern, @Kixunil. I don't know if you've tried running Umbrel for yourself, but we make it very clear during the onboarding that Umbrel is in beta, it should not be considered secure, and you shouldn't put more funds on it than you can afford to lose.

![image](https://user-images.githubusercontent.com/10330103/103434630-3ebf6480-4c2a-11eb-9c40-1af0cb64bac2.png)

Our target audience isn't technical, and most of them don't even know what "SSH" means. We don't want to turn them off with things they don't understand and make them feel like they need a certain level of technical literacy to understand and use the product, as that would end up defeating our entire purpose.

It's also unnecessary for them to understand every single security tradeoff that we've currently made during the beta, as long as they are generally aware of the fact that it is not secure and they shouldn't put anything on it that they can't afford to lose — which we think the final screen of our onboarding has done a good job at.

Kixunil:
Hmm, yet, those people still install it, so I'm not convinced it's sufficient. It also wastes their time if they only learn about it at the last step and may feel pressure of sunken cost fallacy to continue using it despite the warning.

If they shouldn't know what SSH means, perhaps disable it by default instead of using default password? That'd help a lot.

AaronDewes:
> Hmm, yet, those people still install it, so I'm not convinced it's sufficient. It also wastes their time if they only learn about it at the last step and may feel pressure of sunken cost fallacy to continue using it despite the warning.
> 
> If they shouldn't know what SSH means, perhaps disable it by default instead of using default password? That'd help a lot.

We need SSH for debuggin, and we'd have to implement an option to enable it first.

Kixunil:
My understanding is that you use a modified RPI image. If that's the case you just need to touch `/boot/ssh` after flashing the SD card.

Alternatively, you can create two builds. That's how I do it - never release debug/testing versions to general public.

AaronDewes:
[...]

PR #429, opened 2021-01-17
Allow to change SSH password

AaronDewes:
This is just the host part, but it's a first step.

Fixes: Twitter users who will tweet about this issue until a fix is released, ﻿ #381.


AaronDewes:
@lukechilds Do you a know a way to pass the password encrypted? The password will be on the filesystem for a few seconds with this, and I'm not sure if you're comfortable with that, especially if the SSH password should be the same as the user password, but if this amount of time is after the login, an attacker with shell access could already modify lnd from CLI because it's unlocked

lukechilds:
There's no need to pass the password encrypted, if we do then we need to pass the decryption key along too or use a hardcoded one. Either way, if an attacker has access to the environment to view the encrypted password it would be trivial to decrypt.

louneskmt:
I don't think this is a good way to communicate the new password to the system, as it may contain special characters, and file names cannot contain some of them.

AaronDewes:
>ou're right, I'll put it into the file

events/triggers/passwd
```diff
@@ -0,0 +1,13 @@
...
+signal="${1}"
+
+# Try to load Umbrel OS version info to detect Umbrel OS
+[[ -f "/etc/default/umbrel" ]] && source "/etc/default/umbrel"
+
+
+# Only change password on Umbrel OS
+if [[ ! -z "${UMBREL_OS:-}" ]]; then
+    # Script runs as root, we need to set the username
+    echo ${1} | passwd umbrel --stdin
```
Kixunil, line 12: Did you mean to write `cat "$signal" | passwd umbrel --stdin`?

I don't know much about Umbrel but it seems to me you're storing this inside the file not in the file name?

events/triggers/passwd
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+signal="${1}"
```
Kixunil, line 3: Passing password on command line is not great but not too bad given current assumptions. Make sure to add this to your TODO list of things to fix when implementing isolation.

karen
```diff
@@ -63,6 +63,14 @@ fswatch -0 --event=PlatformSpecific $signal_dir | while read -d "" event; do
     args="${signal#$app_prefix}"
   fi
 
+  pw_prefix="passwd-"
+  if [[ "$signal" == "$pw_prefix"* ]]; then
+    trigger="$trigger_dir/passwd"
+    args="${signal#$pw_prefix}"
+    # Delete signal to remove pw from file system
+    rm "$signal_dir/$signal"
```
Kixunil, line 71: If you're worried about it being stored on disk then I recommend using `/dev/shm/` for this, which is stored in RAM and will be definitely wiped at worst after reboot. If you need some file in directory, then symlink will solve this.

Kixunil (commented):
Doesn't seem entirely correct to me. Do you have your events/triggers documented somewhere so that I could help by reviewing it more properly?

events/triggers/passwd
AaronDewes: I already said that I will change it.

karen
AaronDewes: I think the current solution is okay.

Kixunil: Yes, not a big deal at this stage of development.

lukechilds:
Thanks for the review @Kixunil.

>Doesn't seem entirely correct to me. Do you have your events/triggers documented somewhere so that I could help by reviewing it more properly?

We don't have it documented, it's basically just when `events/signals/foo` is touched `events/trigger/foo` is executed. There was some special handling for apps. This was just to quickly allow Docker containers to run certain jobs on the host. A proper RPC server is planned long term.

@AaronDewes the way we usually pass data around is in status files. So we should probably have some sort of `status/password-updated` file which by default is `false`. We then write the password to that file, touch `events/signals/change-password` which causes `events/triggers/change-password` to be executed which reads `status/password-updated` and updates the Umbrel user password with the value, then sets the value of `status/password-updated` to `true`.

This actually introduces a new vulnerability where if an attacker can somehow execute code on the host as a non-umbrel non-root user they can escalate their privileges to umbrel (and then root via sudo) by writing to the signal/status files to set the umbrel password to something they know. This is not an issue right now since it's still better than the current setup, docker provides (some) isolation from compromised services being able to execute on the host so this attack would not at all be easy to pull off. This can be resolved properly with the new RPC server by making sure it only responds to requests with a valid token and tokens are only readable by select users/containers with the required permissions.

Long term this will all be handled at setup and future password changes will be handled more cleanly by an RPC server running on the host without need for writing persistent files. The RPC server will only responds to requests with a valid token and tokens will only be readable by select users/containers with the required permissions.

AaronDewes:
I hope this is better now (the host part, will implement the manager part later)

lukechilds (approved):

lukechilds:
Thanks @AaronDewes!

AaronDewes:
I forgot about htis PR, will implement the manatger part today

PR #434, opened 2021-01-18
Update thunderhub app to use Tor proxy

louneskmt:
As discussed in #402.


lukechilds (approved):
Legend, thanks @louneskmt!

PR #439, opened 2021-01-18
Fix unattended-upgrades package installation

lukechilds:
The package name is `unattended-upgrade` not `unattended-update`.

This was causing an error in the update script causing it to fail.


AaronDewes:
Sorry, should've tested.

lukechilds:
@AaronDewes no worries, me too!

PR #445, opened 2021-01-20
Fix "empty macaroon data" for loop in ride-the-lightning app

### LINKED ISSUE #443: [Ride the Lightning] ERROR: empty macaroon data
I installed ride-the-lightning (RTL) via the app store and wanted to try out the loop feature.
Under lightning -> loop I saw the following error:
Screen Shot 2021-01-19 at 9 40 15 PM
After a bit of investigation I found out that RTL needs a `node.authentication` key in the `RTL-Config.json` file.
However, after exec'ing into the `ride-the-lightning_web_1` docker container, I saw that this key isn't there:
```bash
docker exec -it ride-the-lightning_web_1 sh
/RTL $ cat /data/RTL-Config.json
{
  "defaultNodeIndex": 1,
  "SSO": {
    "rtlSSO": 0,
    "rtlCookiePath": "",
    "logoutRedirectLink": ""
  },
  "nodes": [
    {
      "index": 1,
      "lnNode": "Umbrel",
      "Settings": {
        "userPersona": "MERCHANT",
        "themeMode": "DAY",
        "themeColor": "PURPLE",
        "enableLogging": true,
        "fiatConversion": true
      }
    }
  ],
  "multiPassHashed": "<truncated>"
}
```

According to the [RTL readme](https://github.com/Ride-The-Lightning/RTL#prep-for-execution) it sounds like the `node.authentication` key should look something like this: 
```json
    "Authentication": {
        "macaroonPath": "<Complete path of the folder containing admin.macaroon for the node # 1>",
        "swapMacaroonPath": "<Complete path of the folder containing loop.macaroon for the node # 1>"
      }
```

I found the `admin.macaroon` file under this path inside the RTL web container: `/lnd/data/chain/bitcoin/testnet/admin.macaroon`.
But the `loop.macaroon` file was nowhere to be found. It looks like the docker-compose file for the app is supposed to configure the `swapMacaroonPath` [here](https://github.com/getumbrel/umbrel/blob/master/apps/ride-the-lightning/docker-compose.yml#L36). The path there ultimately resolves to [this path](https://github.com/getumbrel/umbrel/tree/master/apps/ride-the-lightning/rtl) of the repository, and there is no `loops` direcotry.

Hope this analysis helps to quickly resolve the error about missing macaroon data!

louneskmt:
Fixes #443.

### DIFF
--- a/apps/ride-the-lightning/docker-compose.yml
+++ b/apps/ride-the-lightning/docker-compose.yml
@@ -17,6 +17,7 @@ services:
       - "$APP_RIDE_THE_LIGHTNING_PORT:$APP_RIDE_THE_LIGHTNING_PORT"
     volumes:
       - ${APP_DATA_DIR}/rtl:/data
+      - ${APP_DATA_DIR}/loop:/loop
       - ${LND_DATA_DIR}:/lnd:ro
       - ${BITCOIN_DATA_DIR}:/bitcoin:ro
     environment:
@@ -33,7 +34,7 @@ services:
 
         # Loop
         SWAP_SERVER_URL: "https://loop:8081"
-        SWAP_MACAROON_PATH: "/data/loop/$BITCOIN_NETWORK"
+        SWAP_MACAROON_PATH: "/loop/$BITCOIN_NETWORK"
     networks:
       default:
         ipv4_address: $APP_RIDE_THE_LIGHTNING_IP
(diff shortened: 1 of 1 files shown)


louneskmt:
Can somebody test this on a mainnet node? I cannot access mine at the moment.

AaronDewes (dismissed):

louneskmt:
Tested and working on both testnet and mainnet.

lukechilds:
This just needed the correct path/volume passed in to RTL.

I've reverted the change of Loop path because it'll end up losing anyone's loop DDB if they've fixed themselves and also it ends up with the old `.loop` dir inside the new `.loop` dir which could cause problems.

lukechilds (approved):

PR #471, opened 2021-01-29
Use new pattern for configure script

louneskmt:
This PR globalizes the use of the new pattern for setting values in config files templates.


bguillaumat (dismissed):

scripts/configure
```diff
@@ -243,6 +209,8 @@ fi
 # TODO: Update all the above code to use this simpler logic
 for template in "${NGINX_CONF_FILE}" "${BITCOIN_CONF_FILE}" "${LND_CONF_FILE}" "${TOR_CONF_FILE}" "${ELECTRS_CONF_FILE}" "${ENV_FILE}"; do
   # Umbrel
+  sed -i "s/<umbrel-version>/${UMBREL_VERSION}.local/g" "${template}"
+  sed -i "s/<hostname>/${DEVICE_HOSTNAME}/g" "${template}"
```
bguillaumat, line 213: This one shouldn't be with  `.local` after?

louneskmt: Damn good spot, I added it on the wrong line

PR #476, opened 2021-01-31
Update lncm/lnd to 0.12.0

nolim1t:
Update LND to 0.12.0 

Thanks @AaronDewes for reminding about it

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -50,7 +50,7 @@ services:
                         ipv4_address: $BITCOIN_IP
         lnd:
                 container_name: lnd
-                image: lncm/lnd:v0.11.1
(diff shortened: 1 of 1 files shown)


lukechilds:
Thanks @nolim1t!

Are you able to remove the checksum to keep inline with the atomic commit philosophy?

A follow up PR to pin all Docker images would be very welcome!

lukechilds (changes requested):
Thanks @nolim1t!

Are you able to remove the checksum to keep inline with the atomic commit philosophy?

A follow up PR to pin all Docker images would be very welcome!

nolim1t:
Ok done

lukechilds (approved):
Legend!

PR #496, opened 2021-02-09
Remove duplicated depends_on

AaronDewes:
This is not in this PR, it's just my opinion:

In my opinion it would make more sense to have manager & middleware depend on BTC & LND instead of making it the other way around, is there any reason for not doing that? Also, I think lnd should depend on bitcoind

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -136,7 +136,6 @@ services:
                 depends_on: [ manager, bitcoin, lnd ]
                 command: ["./wait-for-node-manager.sh", $MANAGER_IP, "npm", "start"]
                 restart: on-failure
-                depends_on: [ manager ]
                 volumes:
                         - ${PWD}/lnd:/lnd
                         - jwt-public-key:/jwt-public-key


lukechilds (approved):
Great spot, thanks @AaronDewes!

lukechilds:
>In my opinion it would make more sense to have manager & middleware depend on BTC & LND instead of making it the other way around, is there any reason for not doing that?

We have manager as the first dep (apart from Tor) to ensure it is the last to be shutdown. So during OTA updates when we stop all services and then restart the new one, the manager service is down for the shortest time possible. This is important because it's what's responding to the update progress API requests the user sees.

>Also, I think lnd should depend on bitcoind

Strictly speaking, yeah it probably should. However I'm not sure it would make much of a difference in reality since Docker just waits for the process to start, it doesn't wait for the port to be available. So either way, with or without `depends`, LND will start before it can connect to Bitcoin Core. So we may as well get LND started as early as possible so it's ready quicker.

Also technically LND only depends on Core after IBD. Before IBD it's running in Neutrino mode so is not dependant on Core at all.

PR #501, opened 2021-02-10
Add BIND_ADDRESS environment variable

louneskmt:
This PR follows a discussion on Keybase with @lukechilds.

Keeping in mind that it is **not recommended** to run Umbrel on a VPS as it is designed to be run on a secure local network, this PR allows VPS users to bind containers port to a custom IP address (`127.0.0.1` recommended) instead of `0.0.0.0` (which opens Umbrel's ports to the world on a VPS connected directly to internet as there is no router).


scripts/start
```diff
@@ -38,7 +38,7 @@ fi
 
 # Configure Umbrel if it isn't already configured
 if [[ ! -f "${UMBREL_ROOT}/statuses/configured" ]]; then
-  NETWORK="${NETWORK:-mainnet}" "${UMBREL_ROOT}/scripts/configure"
+  NETWORK="${NETWORK:-mainnet}" BIND_ADDRESS="${BIND_ADDRESS:-0.0.0.0}" "${UMBREL_ROOT}/scripts/configure"
```
lukechilds, line 41: You're using `BIND_ADDRESS` here but asking for `UMBREL_BIND_ADDRESS` in the docs:

```
sudo UMBREL_BIND_ADDRESS=127.0.0.1 ./scripts/start
```

scripts/configure
```diff
@@ -102,6 +102,7 @@ ENV_FILE="./templates/.env"
 [[ ! -z ${PREV_ENV_FILE+x} ]] && [[ -f "${PREV_ENV_FILE}" ]] && source "${PREV_ENV_FILE}"
 
 # Umbrel
+BIND_ADDRESS="${UMBREL_BIND_ADDRESS:-BIND_ADDRESS:-0.0.0.0}"
```
lukechilds, line 105: I think to reduce confusion and chance of bugs just stick to a single variable name instead of checking for `UMBREL_BIND_ADDRESS` externally and using `BIND_ADDRESS` internally.

README.md
```diff
@@ -59,6 +59,11 @@ sudo NETWORK=testnet ./scripts/start
 sudo NETWORK=regtest ./scripts/start
 ```
 
+If you host Umbrel on a VPS (while it is **not recommended** as Umbrel is designed to be run on a secure local network), set the `UMBREL_BIND_ADDRESS` environment variable to `127.0.0.1`:
```
lukechilds, line 62: I think it's best to not even mention VPS usage here as we really don't want to encourage it at this point.

We can just say if you don't want to open any ports and only want to expose Umbrel via Tor hidden services then set `UMBREL_BIND_ADDRESS=127.0.0.1`.

lukechilds (changes requested):
Some tiny tweaks needed.

scripts/configure
louneskmt: The thing is that the `BIND_ADDRESS` env var specified by the user is overwritten by the previous run value, as the env file is sourced just above. So once it is set for the first time, it can no longer be changed. 

This is the workaround I found. Can you think of any other?

README.md
louneskmt: Sure.

scripts/configure
lukechilds: Ahh I see. I wouldn't worry about that in this PR. I think the `BITCOIN_NETWORK` and `NGINX_PORT` environment variables have the same issue.

But yeah, that behaviour is annoying and potentially dangerous. Although maybe we should fix that for everything in a follow up PR. We could just save them in a temp var directly before sourcing, and then overwrite them directly after sourcing.

louneskmt: Make sense. I will make the changes 👌

louneskmt: Resolved in 85ecd29.

README.md
louneskmt: Resolved in e3e25b8.

louneskmt:
This has been tested and works well 👌

AaronDewes (approved):
#625, #566 and #385 need to be updated if this is merged.

PR #503, opened 2021-02-11
Generate tor hidden services on configuration

### LINKED ISSUE #304: Can't login via Tor browser
Documenting this until it's fixed. Some users can't get past the login screen on a Tor-enabled browser (login button fades in and out endlessly). 

Restarting Umbrel via the settings page (accessed via the local network on `http://umbrel.local/settings` or `http://<ip-address>/settings`) permanently fixes the issue and prevents it from happening again.

The reason why this happens is that on the first Umbrel run, dashboard's hidden service file doesn't exist, and thus the hidden service URL is not whitelisted by manager/middleware APIs for CORS. We can fix it by starting the manager/middleware containers after checking for existence of the hidden service file using a custom start script.

mayankchhabra:
Fixes #304.

### DIFF
--- a/scripts/configure
+++ b/scripts/configure
@@ -319,6 +319,18 @@ mv -f "$TOR_CONF_FILE" "./tor/torrc"
 mv -f "$ELECTRS_CONF_FILE" "./electrs/electrs.toml"
 mv -f "$ENV_FILE" "./.env"
 
+##########################################################
+######### Generate hidden services on first run ##########
+##########################################################
+
+# Prevents CORS issue if hidden service doesn't exist 
+# on the first run
+if [[ ! -f "${STATUS_DIR}/configured" ]]; then
+  echo "Generating hidden services..."
+  echo
+  docker-compose up --detach tor
+  docker-compose down
+fi
 
 ##########################################################
 ################ Configuration complete ##################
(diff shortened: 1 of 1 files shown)


lukechilds:
There's no guarantee this will create all the hidden service files before it gets killed.

Even if you've tested it and it works, you're relying on the chance the files are created before the process gets killed, it could still fail on slower machines or under different circumstances.

However if you just want to get this out as a quick fix maybe just adding `sleep 1` in between starting and stopping the container is enough to make the failure rate very low.

Alternatively we could loop and check for the existence of the Tor hidden service hostname file, although maybe that's overkill

```bash
wait_for_tor=10
while [[ ! -f "${DASHBOARD_HIDDEN_SERVICE_FILE}" ]]; do
  if [[ "${wait_for_tor}" == 0 ]]; then
    echo "Tor hidden service file wasn't created"
    break
  fi
  sleep 1
  ((wait_for_tor--))
done
```

mayankchhabra:
Ah, that makes sense. I thought the same but could never really reproduce the issue as it almost instantly generated the files.

> Alternatively we could loop and check for the existence of the Tor hidden service hostname file, although maybe that's overkill

Not anymore since you've already written code for it haha. Ok, let's go with it then! Will update.

lukechilds (approved):
Beautiful!

PR #511, opened 2021-02-12
Expose Bitcoin Core RPC port to the local network

### LINKED ISSUE #506: Please make Bitcoin RPC port available locally
Currently docker exposes only port 8333 to local LAN. 
If RPC port 8332 would be exposed as well, we could configure Ledger Live to use the Umbrel node.

lukechilds:
This exposes the RPC port to the local network to allow connections to wallets like Specter, Caravan, Ledger Live, Lily Wallet and Nunchuk without requiring Tor.

RPC is not really designed to be exposed like this, it's a potential DoS attack vector. However this is not an issue during beta while we make the assumption that the local network is secure.

Before a stable release we should have an auth proxy sat in front of Bitcoin Core RPC that is designed to be exposed and is resistant to DoS attacks.

Resolves #506

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -44,6 +44,7 @@ services:
                 restart: on-failure
                 stop_grace_period: 15m30s
                 ports:
+                    - "$BITCOIN_RPC_PORT:$BITCOIN_RPC_PORT"
                     - "$BITCOIN_P2P_PORT:$BITCOIN_P2P_PORT"
                 networks:
                     default:


lukechilds:
After some more consideration I don't think we should enable this: https://github.com/getumbrel/umbrel/issues/506#issuecomment-778223917

unclemantis:
If I want to mine another first level token that piggy backs bitcoin will this be OK? Example: stacks.

PR #517, opened 2021-02-13
Pass Umbrel seed into app script during updates

louneskmt:
This PR makes sure we use the seed from the real Umbrel installation in the entropy generator function during an OTA update.

@lukechilds

### DIFF
--- a/scripts/app
+++ b/scripts/app
@@ -37,8 +37,15 @@ list_installed_apps() {
 
 # Deterministically derives 128 bits of cryptographically secure entropy
 derive_entropy () {
+  # Make sure we use the seed from the real Umbrel installation if this is
+  # an OTA update.
+  SEED_FILE="${UMBREL_ROOT}/db/umbrel-seed/seed"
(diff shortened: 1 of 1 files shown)


scripts/app
```diff
@@ -37,8 +37,15 @@ list_installed_apps() {
 
 # Deterministically derives 128 bits of cryptographically secure entropy
 derive_entropy () {
+  # Make sure we use the seed from the real Umbrel installation if this is
+  # an OTA update.
+  SEED_FILE="${UMBREL_ROOT}/db/umbrel-seed/seed"
+  if [[ -f "${UMBREL_ROOT}/../.umbrel" ]]; then
```
lukechilds, line 43: Are you able to just add a sanity check the the current `SEED_FILE` is missing here too.

lukechilds (changes requested):

scripts/app
lukechilds:
```suggestion
  if [[ ! -f "${SEED_FILE}" ]] && [[ -f "${UMBREL_ROOT}/../.umbrel" ]]; then
```

PR #518, opened 2021-02-13
Add feeurl to lnd.conf

mayankchhabra:
LND v0.12.0 [doesn't work](https://github.com/lightningnetwork/lnd/blob/4a96c2e55ab356e45aaa1904601d04a7beb703b0/sample-lnd.conf#L222-L225) without `feeurl` when using neutrino on mainnet:

```
lnd | unable to create chain control: --feeurl parameter required when running neutrino on mainnet
```

Since we don't have our own fee estimation endpoint, we can use their default. Updated [neutrino-switcher](https://github.com/getumbrel/docker-lnd-neutrino-switch) to remove this setting once LND has switched to `bitcoind`: https://github.com/getumbrel/docker-lnd-neutrino-switch/pull/2.


templates/lnd-sample.conf
```diff
@@ -9,6 +9,7 @@ tlsextraip=<lnd-ip>
 tlsextradomain=<hostname>
 tlsautorefresh=1
 tlsdisableautofill=1
+feeurl=https://nodes.lightning.computer/fees/v1/btc-fee-estimates.json
```
lukechilds, line 12: Shouldn't this only be added on mainnet?

lukechilds: The API includes block hash, I wonder if that could cause issues if there's a mismatch:

```json
{
  "current_block_hash": "0000000000000000000613290e0bd5ee22838eb1079f93ba178835fdd8ce117d",
  "fee_by_block_target": {
    "2": 164820,
    "3": 117088,
    "7": 106156,
    "13": 100904,
    "14": 90984,
    "16": 87284,
    "18": 83064,
    "19": 82964,
    "25": 61776,
    "37": 53184,
    "49": 50748,
    "61": 34452,
    "73": 17292,
    "85": 9092,
    "97": 8036,
    "98": 8032,
    "121": 7112,
    "123": 6996,
    "130": 6016
  }
}
```

Or if it ignores the hash and just uses high fees for regtest/testnet then that's fine.

I'll test on regtest now.

lukechilds: Ahh can't test cos no Neutrino on regtest.

It might end up breaking transactions on testnet or just making them high fee. I don't have a testnet instance on hand to test with right now. Let's merge and test after this release since it's only going to affect testnet.

lukechilds: Actually this does break transactions.

On regtest the fee slider gets greyed out and I can't send any transactions.

In the LND logs I see:

```
[ERR] RPCS: [/lnrpc.Lightning/EstimateFee]: unable to query fee estimator: conf target of 1 is too low, minimum accepted is 2
```

Everything works again if I remove this line from the config.

Still, fine to merge for this release since it only affects regtest/testnet environments. We need to fix right after though otherwise we can't send funds in dev environments.

PR #545, opened 2021-02-17
Upload to umbrel-paste.vercel.app in debug script

AaronDewes:
Please note that on shorter, one line strings, you'll get an error "Content is too short", but not with the debug script (tested on a manual install, not Umbrel OS and not containers running for shortest possible output)


lukechilds (approved):

lukechilds:
Awesome stuff @AaronDewes, working great for me!

PR #562, opened 2021-02-23
Allow user to set NGINX_PORT

junderw:
I can't use port 80, so I have made it configurable.

As long as no internal service relies on the main port being 80 (which doesn't make much sense) this should be a harmless change that will help people fiddling with it on AWS etc etc.

Usage:
`sudo UMBREL_LISTEN_PORT=12345 ./scripts/start`

Question:

How will this be affected by OTA update? If I set a port, then update, with the ENV stick?


lukechilds:
Awesome thanks for this @junderw!

>How will this be affected by OTA update? If I set a port, then update, with the ENV stick?

No you'll lose it, but we do have a way of dealing with this. You can persist settings in the `.env` file by adding a placeholder entry in [`templates/.env-sample`](https://github.com/getumbrel/umbrel/blob/master/templates/.env-sample).

Like:

```
UMBREL_LISTEN_PORT=<umbrel-listen-port>
```

Then in the configure script ([`scripts/configure`](https://github.com/getumbrel/umbrel/blob/master/scripts/configure)) which runs the first time Umbrel is started and also during each OTA update you need to check if the value is set, and if not, set it to a default, then write it to the `.env` file. The configure script is essentially a pure function that takes in a `.env` and spits out all the config files.

So you'd add something to the configure script to set the default like:

```bash
UMBREL_LISTEN_PORT=${UMBREL_LISTEN_PORT:-80}
```

and then in [this block](https://github.com/getumbrel/umbrel/blob/721bb1d463346683a5b58ceab586afc323cb2ed2/scripts/configure#L265) write it to the config file with:

```bash
sed -i "s/<umbrel-listen-port>/${UMBREL_LISTEN_PORT}/g" "${template}"
```

I know this code is a bit repetitive, we need to refactor it, but it does the job for now!

Once you've added that you can test an OTA update by running:

```
sudo scripts/update/update --repo getumbrel/umbrel#v0.3.5
```

Replace v0.3.5 with the version you're currently running it Umbrel will run through the OTA update process and just update to the same version you're already running. If everything works well the port should be persisted in `.env` after OTA and will be used when Umbrel is restarted.

lukechilds:
Sorry that should be:
```
sudo scripts/update/update --repo junderw/umbrel#changeListenPort
```

to OTA update to this PR branch otherwise it'll clear all your changes 🤦‍♂️.

junderw:
Verified 0b039a8649fb472d89eb94ce088723b6fb4e340e

It sticks beyond updates.

1. Initially can be set just like NETWORK. passing after sudo for the start script
2. After update etc. since the configure script will run without any env, it will print 80 into `.env`
3. If someone manually goes into `.env` after initial startup and changes the value then restarts using stop and start scripts, the new port is used.

lukechilds:
Thank you @junderw!

I have a few other things I need to focus on this week but I'll get this tested and merged before the next release.

louneskmt:
Duplicate https://github.com/getumbrel/umbrel/pull/256.

junderw:
Close in favor of #256

louneskmt:
@junderw You can review #256 and propose the changes that were here 😁

junderw:
Reopening at the request of @lukechilds

lukechilds (approved):
Thanks for this @junderw!

Tested and working great.

Also switched to @nolim1t's proposed variable name and gave him contribution credit.

PR #566, opened 2021-02-24
Add Lndhub app

AaronDewes:
# App Submission

### App name

LndHub

### Version

1.3.2

Allows to connect a BlueWallet to your Umbrel.

### Summary of the app

Wrapper for Lightning Network Daemon. It provides separate accounts and trust minimization for end users.

### Developed by
BlueWallet

### Developer website
https://bluewallet.io/

### Source code repository

https://github.com/BlueWallet/LndHub

### Support link
https://t.me/bluewallet

### Uses
- [ ] Bitcoin Core
- [ ] Electrum server
- [x] LND

### 256x256 SVG icon & App screenshots

https://github.com/getumbrel/umbrel-apps-gallery/pull/6

### I have tested my app on:
- [x] [Custom Umbrel install on Linux](https://github.com/getumbrel/umbrel#-installation)


apps/lnbits/docker-compose.yml
```diff
@@ -34,7 +34,7 @@ services:
       # App
       LNBITS_SITE_TITLE: "LNbits - Umbrel"
       LNBITS_DEFAULT_WALLET_NAME: "LNbits wallet"
-      LNBITS_DISABLED_EXTENSIONS: "amilk"
+      LNBITS_DISABLED_EXTENSIONS: "amilk,lndhub"
```
louneskmt, line 37: I think you shouldn't disable LNDHub here. Because the whole point of LNbits is offer the ability to have multiple users on one node. Each wallet / user has its own LNDHub and thus can connect their BlueWallet to it, while the LNDHub app is for only one user (= the node wallet). 

At least it is what I've understood, correct me if I'm wrong.

apps/lndhub/docker-compose.yml
```diff
@@ -0,0 +1,44 @@
...
+services:
+        redis:
+              container_name: redis
+              image: redis:6.2-alpine
+              command: redis-server --requirepass moneyprintergobrrr
+              restart: on-failure
+              stop_grace_period: 5m
+              stop_signal: SIGINT
+              ports:
+                  - "$APP_LNDHUB_REDIS_PORT:$APP_LNDHUB_REDIS_PORT"
```
louneskmt, line 18: No need to expose the Redis port here, as it's only used by the other container.

apps/lndhub/docker-compose.yml
```diff
@@ -0,0 +1,44 @@
...
+                ports:
+                    - "${APP_LNDHUB_PORT}:${APP_LNDHUB_PORT}"
+                volumes:
+                        - ${LND_DATA_DIR}:/lnd:ro
+                environment:
+                    PORT: "${APP_LNDHUB_PORT}"
+                    TOR_URL: "$APP_HIDDEN_SERVICE"
+                    LND_CERT_FILE: "/lnd/tls.cert"
+                    LND_ADMIN_MACAROON_FILE: "/lnd/data/chain/bitcoin/${BITCOIN_NETWORK}/admin.macaroon"
+                    CONFIG: '{ "redis": { "port": $APP_LNDHUB_REDIS_PORT, "host": "$APP_LNDHUB_REDIS_IP", "family": 4, "password": "moneyprintergobrrr", "db": 0 }, "lnd": { "url": "$LND_IP:$LND_GRPC_PORT", "password": ""}}'
```
louneskmt, line 41: Re my previous comment: you can hardcode the redis port here (and use the hostname "redis" instead of the IP address, but it also works like this).

apps/lnbits/docker-compose.yml
Overtorment: its exactly the opposite. lnbits just replicated lndhub. poorly.

apps/registry.json
```diff
@@ -232,5 +232,28 @@
...
         "path": "",
         "defaultPassword": ""
+    },
+    {
+        "id": "lndhub",
+        "category": "Wallet Servers",
+        "name": "BlueWallet Server",
+        "version": "1.3.0",
+        "tagline": "Connect a BlueWallet to your Umbrel",
+        "description": "Wrapper for Lightning Network Daemon. It provides separate accounts and trust minimization for end users.",
```
Overtorment, line 242: "bla bla onboard your family and friends to Lightning powered by your own node"

cc @ncoelho

Overtorment (dismissed):
lgtm

apps/lnbits/docker-compose.yml
AaronDewes: Yeah, LndHub offers multiple accounts too, and each LNbits user having multiple sub-acoounts would be a bit to complicated.

lukechilds: @Overtorment are there known issues with LNbits LNDhub implementation?

I think there is a slightly different use case between BlueWallet Server/LNbits+LNDhub here.

You can create sub accounts in LNbits for things like earnings from a paywall link forwarder. Then you can add that LN wallet to BlueWallet to track your earnings from the paywall. You could also then easily sweep funds to your main wallet via BlueWallet without having to access the LNbits web wallet etc.

I see it more as a way to just interface with the LNbits wallets via BlueWallet rather than a serious LNDhub replacement.

So:
- LNbits+LNDhub: Access your LNbits wallets from BlueWallet.
- BlueWallet Server: Back your BlueWallet with your own Lightning node.

What do you think?

ncoelho: Thing here is "lndhub" should never be visible to the user, thus the confusion, there's nothing people can do there. Should just be a kid of "BlueWallet Server" as a way to connect to the app. If lnbits/lndhub connection won't break lndhub, all is good.

templates/torrc-sample
```diff
@@ -76,4 +76,8 @@ HiddenServicePort 80 <app-samourai-server-ip>:80
 HiddenServiceDir /var/lib/tor/app-samourai-server-whirlpool
 HiddenServicePort 80 <app-samourai-server-whirlpool-ip>:<app-samourai-server-whirlpool-port>
 
+# LndHub Hidden Service
+HiddenServiceDir /var/lib/tor/app-lndhub
+HiddenServicePort 3000 <app-lndhub-ip>:<app-lndhub-port>
```
junderw, line 81: Is it required that the user connect via `http://xxxx.onion:3000/` when using tor?

All the other apps use 80 to eliminate the port from the URL, but does LNDHub need it?

junderw:
Any blockers for this? I personally kinda want to use this.

I'm guessing it needs some screenshots?

templates/torrc-sample
Overtorment: if thats a question about bw - bw can use any port

louneskmt: Nope @Overtorment, it's not about BW internal port, but the port used for the app Tor hidden service. 

Usually, we set it to 80 so the onion hostname is sufficient to connect, but here it is set to 3000, so the user would need to enter blabla.onion:3000 instead of blabla.onion.

maltokyo:
I see Bluewallet and LNDHub now finalized TOR inclusion.  @AaronDewes do you think this now can finalise?  would be cool to see.  As always, I can test as needed, please share if anything.

AaronDewes:
Yeah, I'm making this ready now.

maltokyo:
Awesome!

apps/lndhub/docker-compose.yml
```diff
@@ -0,0 +1,35 @@
...
+              stop_grace_period: 5m
+              stop_signal: SIGINT
+              volumes:
+                  - ${APP_DATA_DIR}/redis:/data
+              networks:
+                  default:
+                      ipv4_address: $APP_LNDHUB_REDIS_IP
+        web:
+                container_name: web
+                image: bluewalletorganization/lndhub:v1.3.0
```
lukechilds, line 18: This image is arm64 only: https://hub.docker.com/layers/bluewalletorganization/lndhub/v1.3.0/images/sha256-2e0785e93ff328f1b33521573e5725606c10d8654c9c6dd0a557469890a1c398?context=explore

Umbrel needs multi-arch builds supporting at least arm64 and amd64.

lukechilds: Looks like both archs are being built but they're being pushed over the same tag: https://github.com/BlueWallet/LndHub/blob/f64a8c8102b19e87ea3f9f877e7ba27976dcc304/.github/workflows/tag.yml

@AaronDewes you need to do a single buildx that builds both archs instead of doing two buildx runs of separate archs with the same tag.

lukechilds (changes requested):

apps/lndhub/docker-compose.yml
AaronDewes: Sorry about that, I thought multi-arch separate builds were faster.

lukechilds (approved):
Great job @AaronDewes! Working brilliantly.

maltokyo:
Really great, thanks @AaronDewes and @lukechilds !

I have been testing, and did you know (I didn't) that even when people do a refill into their Bluewallet Lightning wallet, it doesn't actually open any channels for that refill and leaves the funds on-chain?

How will Umbrel then make channels with LNDHub user funds, is there some automated way that is being handled, or an admin needs to go in and do it manually?

You can see my discussion with @Overtorment here, not complaining, I just didn't know this, so leaving here in case it is the same for others:
https://github.com/BlueWallet/BlueWallet/issues/3051

junderw:
This should be mentioned somewhere on the app store imo.

A lot of questions will pop up under assumption that off-chain deposit on Bluewallet equals open channel. Since they hide that UX-wise.

junderw:
I do not think this is a blocker tbh. Just make sure the App description is clear that "LNDHub will not automatically create channels for you, so if you "deposit to lightning" via BlueWallet and you have no channels all payments will fail."

PR #568, opened 2021-02-25
Update mempool app to v2.1.2

bguillaumat:
```
This is a patch release that optimizes memory usage for embedded devices like Raspberry Pi

Notes:

The disk cache now has its own folder, so mv cache*.json ./cache/ before restart to migrate
Changes:

Optimize memory usage when writing disk cache (#342)
Reduce backend maximum heap size setting to 2G (#345)
Enable mempool clear protection on all backends (#335)
Make mempool clear protection timeout configurable (#343)
Minor tweaks to About page text, links (#350)
Logo design update (#349) by @pedromvpg
Fix style on block hover (#347) by @Czino and @Eric-Machinas
```

### DIFF
--- a/apps/mempool/docker-compose.yml
+++ b/apps/mempool/docker-compose.yml
@@ -8,7 +8,7 @@ x-logging:
 
 services:
   web:
-    image: mempool/frontend:v2.1.1@sha256:17e7fedcd27b6f99de2f159f0f6372b76dd5825f1a98c1b0114ea9564cab1c0e
+    image: mempool/frontend:v2.1.2@sha256:69fefe55dc1eb4c8373c32d57362df7eeb672fc591d55033d68ae79355274dcc
     user: "1000:1000"
     init: true
     logging: *default-logging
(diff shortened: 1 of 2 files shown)


downet:
Is it possible to make also the logging non debug? As I saw it's writing all debug info towards sd card.

apps/mempool/docker-compose.yml
```diff
@@ -44,7 +44,7 @@ services:
       MYSQL_DATABASE: "mempool"
       MYSQL_USER: "mempool"
       MYSQL_PASS: "mempool"
-      CACHE_DIR: "/backend/cache/"
```
lukechilds, line 47: Why is this change needed?

lukechilds:
Thanks for this @bguillaumat!

>The disk cache now has its own folder, so mv cache*.json ./cache/ before restart to migrate

Do we need to do any special handling on our end for this or will it not affect us if we're already passing in an explicit `CACHE_DIR ` path?

bguillaumat:
@lukechilds nah it's all good for us it's an update Simon made for old mempool to be compliant with the cache_dir var

apps/mempool/docker-compose.yml
bguillaumat: They've update mempool so now their code add this var to "/cache.json" so I've removed the useless slash on the docker-compose

lukechilds:
>Is it possible to make also the logging non debug? As I saw it's writing all debug info towards sd card.

@downet I think a better solution is for us to redirect Umbrel logs to the SSD. I'll work on that soon.

apps/mempool/docker-compose.yml
lukechilds: Ahh, good spot!

lukechilds (approved):

lukechilds:
Tested and works perfectly, thanks @bguillaumat!

PR #571, opened 2021-02-26
Update debug script for server v2

AaronDewes:


scripts/debug
```diff
@@ -34,7 +34,7 @@
 
 # function to upload the output to our paste server
 # Based on https://github.com/seejohnrun/haste-client#lightweight-alternative
-upload() { a=$(cat); curl -X POST -s -d "$a" https://umbrel-paste.vercel.app/documents | awk -F '"' '{print "https://umbrel-paste.vercel.app/"$6}'; }
+upload() { a=$(cat); curl -H "Content-Type: text/plain" -X POST -s -d "$a" https://umbrel-paste.vercel.app/documents | awk -F '"' '{print "https://umbrel-paste.vercel.app/"$6}'; }
```
lukechilds, line 37: Regarding the new `-H "Content-Type: text/plain"` requirement. Does this mean users running old versions of Umbrel will no longer be able to upload correctly?

If so maybe we should change the URL so old users upload to the old URL and new users upload to the new URL.

Or will old users still be able to upload without this header?

AaronDewes: It's not a requirement, it just fixes a bug that was also present on the old server where vercel parsed the data as JSON or other formats instead of plain text.

AaronDewes: You can see this paste that a user created today for an example of a 0.3.5 paste, it works fine, I made sure its compatible: https://umbrel-paste.vercel.app/5u00iv0

lukechilds: I'm getting:

```
================
==== Result ====
================
This script could not automatically detect an issue with your Umbrel.
Please share the following links and paste it in the Umbrel Telegram group (https://t.me/getumbrel) so we can help you with your problem.
scripts/debug: line 37: /usr/bin/curl: Argument list too long
```

I think cause we're loading the output into a string var and passing it in as an arg. Maybe it would be better to pipe the output to curl and then tell curl to read from stdin.

AaronDewes: Done

lukechilds: Wow that was quick!

Seems like maybe that breaks newline formatting though, I'm assuming it's not supposed to look like this: https://umbrel-paste.vercel.app/2030e66

AaronDewes: Oh, I actually had that in development, thought it was a server-side bug, but apparently switching the upload function was the cause, I'll fix it

AaronDewes: Just clicked the commit button and wanted to commit exactly what you did now, you were 1sec faster.

AaronDewes: > https://umbrel-paste.vercel.app/2030e66

I just remembered, it was actually both, server-side and client side that ignored newlines, but server is fixed now.

lukechilds: It's ok, fixed, needs --data-binary to respect newlines.

lukechilds: Also, as a general rule, I always prefer to use long hand argument names in scripts, since it results in more readable code. Especially when other people come across it in the future who may be less familiar with CLI options.

Also, if there's a command with lots of args, I like to split them up with newlines for readability too.

e.g:

```bash
upload() { a=$(cat); curl -X POST -s -d "$a" https://umbrel-paste.vercel.app/documents | awk -F '"' '{print "https://umbrel-paste.vercel.app/"$6}'; }
```
vs
```bash
upload() {
  curl \
    --header "Content-Type: text/plain" \
    --request POST \
    --silent \
    --data-binary @- \
    https://umbrel-paste.vercel.app/documents \
    | awk -F '"' '{print "https://umbrel-paste.vercel.app/"$6}'
}
```
does that make sense?

AaronDewes: Yes, I wasn't sure what you prefer, and that was the original code (I copy and pasted the function)

PR #573, opened 2021-02-26
Rename mempool db container to mariadb

### LINKED ISSUE #572: Samourai Server app trying to access Mempool DB
From mempool logs:

```
db_1   | 2021-02-26 10:40:26 6 [Warning] Access denied for user 'samourai'@'10.21.21.25' (using password: YES)
```

It doesn't look like a DB IP is passed in to the Dojo process: https://github.com/getumbrel/umbrel/blob/721bb1d463346683a5b58ceab586afc323cb2ed2/apps/samourai-server/docker-compose.yml#L45-L48

@louneskmt How does it know what to connect to, is it using the `db` hostname?

If so it looks like we're having a collision. Seems like it isn't safe to rely on the service hostnames for networking as we can have cross app collisions. e.g `db` inside the Samourai Server `node` process appears to be resolving to the Mempool `db` IP.

lukechilds:
Resolves #572 

This prevents a hostname collision between the samourai-server_db and mempool_db container.

The samourai-server_node container connects to the DB via the `db` hostname. It was expected that this would always resolve to the IP of the `db` service local to the Docker Compose configuration it was called from. However in some cases it appears that when `db` is requested from samourai-server_node it resolves to the IP of mempool_db instead of samourai-server_db.

Renaming from mempool_db to mempool_mariadb is a quick fix to ensure this won't ever happen. However ideally we would be able to pass in an explicit DB IP to the samourai-server_node container instead of relying on Docker's hostname resolver.

### DIFF
--- a/apps/mempool/docker-compose.yml
+++ b/apps/mempool/docker-compose.yml
@@ -48,7 +48,7 @@ services:
     networks:
        default:
          ipv4_address: $APP_MEMPOOL_API_IP
-  db:
+  mariadb:
     image: mariadb:10.5.8
     user: "1000:1000"
     logging: *default-logging


bguillaumat:
Hey @lukechilds if you change db to mariadb you have to rename the wait-for.sh cmd in the file from db:3306 to mariadb:3306 too

lukechilds:
Oops! I did test everything worked but it must have been running against Samourai Server db. Will fix now, thanks!

PR #580, opened 2021-02-27
Add UAS check

AaronDewes:
Thanks to @louneskmt for helping with this PR.

This PR checks if a drive has UAS issues and if it has, UAS gets disabled.

### DIFF
--- a/scripts/umbrel-os/external-storage/monitor
+++ b/scripts/umbrel-os/external-storage/monitor
@@ -15,8 +15,9 @@ check_if_not_already_running() {
 }
 
 get_uas_string () {
-  ID_VENDOR=$(sudo -u umbrel udevadm test /block/${block_device} | grep "ID_VENDOR_ID" | sed 's/ID_VENDOR_ID=//')
-  ID_MODEL=$(sudo -u umbrel udevadm test /block/${block_device} | grep "ID_MODEL_ID" | sed 's/ID_MODEL_ID=//')
+  UDEVADM_DATA=$(sudo -u umbrel udevadm test /block/${block_device} 2> /dev/null)
+  ID_VENDOR=$(echo "${UDEVADM_DATA}" | grep "ID_VENDOR_ID" | sed 's/ID_VENDOR_ID=//')
+  ID_MODEL=$(echo "${UDEVADM_DATA}" | grep "ID_MODEL_ID" | sed 's/ID_MODEL_ID=//')
   echo "${ID_VENDOR}:${ID_MODEL}:u"
 }
 
--- a/scripts/umbrel-os/external-storage/monitor
+++ b/scripts/umbrel-os/external-storage/monitor
@@ -15,8 +15,8 @@ check_if_not_already_running() {
 }
 
 get_uas_string () {
-  ID_VENDOR=$(udevadm test /block/${block_device} | grep "ID_VENDOR_ID" | sed 's/ID_VENDOR_ID=//')
-  ID_MODEL=$(udevadm test /block/${block_device} | grep "ID_MODEL_ID" | sed 's/ID_MODEL_ID=//')
+  ID_VENDOR=$(sudo -u umbrel udevadm test /block/${block_device} | grep "ID_VENDOR_ID" | sed 's/ID_VENDOR_ID=//')
+  ID_MODEL=$(sudo -u umbrel udevadm test /block/${block_device} | grep "ID_MODEL_ID" | sed 's/ID_MODEL_ID=//')
   echo "${ID_VENDOR}:${ID_MODEL}:u"
 }
 
--- a/scripts/umbrel-os/external-storage/monitor
+++ b/scripts/umbrel-os/external-storage/monitor
@@ -48,7 +48,6 @@ main () {
       echo "External storage might have UAS issues"
       sed "s/usb-storage.quirks=/usb-storage.quirks=$(get_uas_string),/g" -i /boot/cmdline.txt
       echo "Rebooting"
-      docker kill $(docker ps -aq) 
       sudo reboot
     fi
 
--- a/scripts/umbrel-os/external-storage/monitor
+++ b/scripts/umbrel-os/external-storage/monitor
@@ -45,10 +45,10 @@ main () {
       break
     fi
 
-    if [[ $(dmesg | grep --quiet "uas_eh_abort_handler") > 0 ]]; then
+    if [[ $(dmesg) == *"uas_eh_abort_handler"* ]]; then
       echo "External storage might have UAS issues"
       UAS_STRING=$(get_uas_string)
-      if [[ $(cat /boot/cmdline.txt | grep --quiet "${UAS_STRING}") == 0 ]]; then
+      if [[ $(cat /boot/cmdline.txt) == *"${UAS_STRING}"* ]]; then
         sed "s/usb-storage.quirks=/usb-storage.quirks=$(get_uas_string),/g" -i /boot/cmdline.txt
         echo "Rebooting"
         sudo reboot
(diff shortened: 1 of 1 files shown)


louneskmt (dismissed):

scripts/umbrel-os/external-storage/monitor
```diff
@@ -38,6 +44,14 @@ main () {
       break
     fi
 
+    if [[ $(dmesg | grep --quiet "uas_abort_handler") > 0 ]]; then
+      echo "External storage might have UAS issues"
+      sed "s/usb-storage.quirks=/usb-storage.quirks=$(get_uas_string),/g" -i /boot/cmdline.txt
+      echo "Rebooting"
+      docker kill $(docker ps -aq)
```
lukechilds, line 51: Why is this needed?

scripts/umbrel-os/external-storage/monitor
```diff
@@ -38,6 +44,14 @@ main () {
       break
     fi
 
+    if [[ $(dmesg | grep --quiet "uas_abort_handler") > 0 ]]; then
+      echo "External storage might have UAS issues"
+      sed "s/usb-storage.quirks=/usb-storage.quirks=$(get_uas_string),/g" -i /boot/cmdline.txt
```
lukechilds, line 49: Is there ever a situation where `usb-storage.quirks=` could be empty so we end with a comma and break it?

AaronDewes: No, except the user emptied it, but then it's the users fault. This has been in Umbrel OS since v0.1.1, an empty list was never there, when the list doesn't exist at all, sed would do nothing. An empty list would maybe even cause the kernel to refuse boot because there is nothing after the =, but I'm not sure about that

scripts/umbrel-os/external-storage/monitor
AaronDewes: It isn't, I'll remove it.

scripts/umbrel-os/external-storage/monitor
```diff
@@ -14,6 +14,12 @@ check_if_not_already_running() {
   fi
 }
 
+get_uas_string () {
+  ID_VENDOR=$(udevadm test /block/${block_device} | grep "ID_VENDOR_ID" | sed 's/ID_VENDOR_ID=//')
```
lukechilds, line 18: If I run this on either of my Pis:

```
sudo udevadm test /block/sda | grep "ID_VENDOR_ID"
```

I just get loads of stuff printed to stderr and nothing matching `ID_VENDOR_ID`. Is that expected?

AaronDewes: > I just get loads of stuff printed to stderr

Yeah, that's normal, udevadm test is quite verbose.

> and nothing matching ID_VENDOR_ID. 

That isn't normal.

Are you testing this on a Pi with an SSD connected? Because I think it might not work on a normal PC with an internal SSD.

```
ID_VENDOR=$(udevadm test /block/sda | grep "ID_VENDOR_ID" | sed 's/ID_VENDOR_ID=//')
ID_MODEL=$(udevadm test /block/sda | grep "ID_MODEL_ID" | sed 's/ID_MODEL_ID=//')
UAS_RULE="${ID_VENDOR}:${ID_MODEL}:u"
echo $UAS_RULE
```

AaronDewes: Sorry, I didn't read the first sentence

lukechilds: Oh, I was running as root which fails, because this script will be running as root.

I've just tested and running as non-root does indeed work.

Doesn't that mean this script will always fail at boot?

lukechilds:
```
umbrel@umbrel:~/umbrel $ udevadm test /block/sda 2>/dev/null | grep "ID_VENDOR_ID"
ID_VENDOR_ID=04e8
umbrel@umbrel:~/umbrel $ sudo udevadm test /block/sda 2>/dev/null | grep "ID_VENDOR_ID"
umbrel@umbrel:~/umbrel $
```

scripts/umbrel-os/external-storage/monitor
```diff
@@ -38,6 +45,13 @@ main () {
       break
     fi
 
+    if [[ $(dmesg | grep --quiet "uas_abort_handler") > 0 ]]; then
+      echo "External storage might have UAS issues"
+      sed "s/usb-storage.quirks=/usb-storage.quirks=$(get_uas_string),/g" -i /boot/cmdline.txt
+      echo "Rebooting"
+      sudo reboot
```
lukechilds, line 52: We need to be really careful here if we're gonna try to auto patch and reboot. If we mess something up we risk sending the user into a permanent boot loop.

Like if there's some other reason the `uas_abort_handler` string ends up in dmesg or if for whatever reason adding the device to `usb-storage.quirks` doesn't resolve the issue, the user will be stuck in a boot loop.

A simple sanity check to test if `get_uas_string` already exists in `usb-storage.quirks` would go a long way to preventing this. If it already exists just don't do anything and continue to boot.

lukechilds (changes requested):
Has anyone tested this PR out on a Pi with an SSD attached that needs quirks so this logic all gets triggered and tested on a real device?

AaronDewes:
> Has anyone tested this PR out on a Pi with an SSD attached that needs quirks so this logic all gets triggered and tested on a real device?

Not yet, I'll do soon, I got a "broken" SSD

lukechilds:
Just an update in my thinking on this PR. After @AaronDewes [discussion on the mailling list](https://marc.info/?l=linux-usb&m=161504851014681&w=2) it sounds like the issues we've been experiencing are not actually UAS issues but power issues from the Raspberry Pi.

UAS allows drives to do stuff in parallel which gives a performance boost but as a side effect also results in the using more power. Blacklisting UAS means the device falls back to mass-storage mode which doesn't do stuff in parallel so the device ends up consuming less power.

It's not UAS that's causing the issues, it's that the Pi isn't able to reliably power devices in UAS mode, but when running in mass-storage mode they use less power which the Pi can provide so we don't get issues.

I would propose we blanket disable UAS for all devices as a workaround for now. It's not safe to check the logs because even if we don't see that error, some weird combination of high CPU usage + high disk concurrency could push a UAS device over the edge and cause issues at any point.

Ideally we could blanket disable UAS with some kernel param like `uas=false` or `usb-storage.quirks=*`. If that's not possible, it's not worth compiling custom kernels over, we can just change the logic in this PR to:

```
- boot
- get storage device id
- check we've blacklisted it
- if not
- - blacklist it
- - reboot
- if we have
- - continue boot as normal
```

We should have a look into UAS vs mass-storage performance differences to see what kind of perf hit we're gonna take for this. But I guess really it doesn't matter, we can't have devices crashing or experiencing corruption for better performance. Stability is much more important, and most users are already running our recommended hardware in UAS mode and achieving ~3 day IBD which seems reasonable.

Obviously this is a short term solution to resolve stability issues. Long term we need to transition the recommended hardware to something that's able to reliably power common storage devices at full speed (Raspberry Pi 5 🤞).

AaronDewes:
> Ideally we could blanket disable UAS with some kernel param like uas=false or usb-storage.quirks=*. If that's not possible, it's not worth compiling custom kernels over, we can just change the logic in this PR to:

I also asked that on the mailing list, and it's currently not possible.

But I'll change the logic then to just blacklist all devices.

AaronDewes:
> Obviously this is a short term solution to resolve stability issues. Long term we need to transition the recommended hardware to something that's able to reliably power common storage devices at full speed (Raspberry Pi 5 :crossed_fingers:).

Can you read the messages I send on Telegram about the recommended hardware?

Issue #584, opened 2021-02-27
"Not allowed by CORS" on dashboard


vVdLII8WbUien:
I tried using 
```sh
sudo NGINX_PORT=8000 ./scripts/start
```
But got a snackbar error "Not allowed by CORS".

My fix was changing "scripts/start" from:
```sh
DEVICE_IP="$(ip -o route get to 8.8.8.8 | sed -n 's/.*src \([0-9.]\+\).*/\1/p')"
```
to this
```sh
DEVICE_IP="$(ip -o route get to 8.8.8.8 | sed -n 's/.*src \([0-9.]\+\).*/\1/p'):$NGINX_PORT"
```

louneskmt:
From what url are you accessing the dashboard?

vVdLII8WbUien:
http://192.168.178.18:8000/dashboard

phenomeus:
try this: https://github.com/getumbrel/umbrel/issues/398#issuecomment-782348577

phenomeus:
with 0.3.6. and also port 8000 I hadn't to edit this anymore.

MaikB85:
I still need to do the workaround from @vVdLII8WbUien after every update of Umbrel to get it running correctly. Otherwise bitcoin and lnd only show loading animation and no balance is visible.
I run Umbrel on a virtual machine with ubuntu. I need to run it on different port because other software use standard port of Umbrel.

Maybe someone can fix this issue? That would be nice.

nmfretz:
closing old support issue

Issue #398, opened 2021-01-05
not allowd by CORS when logging in on public domain


grmkris:
For easier demoing to others I decided to host umbrel on a vps.  
The vps itself is managed by plesk and is running centos.  

I managed to setup and sucessfully run umbrel. I also connected to it via public address (umbrel.payserver.org)
When trying to login with a password I get the following error:   

```
umbrel-manager
Not allowed by CORS
/v1/account/login
Error: Not allowed by CORS
    at origin (/app/middlewares/cors.js:14:23)
    at /app/node_modules/cors/lib/index.js:219:13
    at optionsCallback (/app/node_modules/cors/lib/index.js:199:9)
    at corsMiddleware (/app/node_modules/cors/lib/index.js:204:7)
    at Layer.handle [as handle_request] (/app/node_modules/express/lib/router/layer.js:95:5)
    at trim_prefix (/app/node_modules/express/lib/router/index.js:317:13)
    at /app/node_modules/express/lib/router/index.js:284:7
    at Function.process_params (/app/node_modules/express/lib/router/index.js:335:12)
    at next (/app/node_modules/express/lib/router/index.js:275:10)
    at expressInit (/app/node_modules/express/lib/middleware/init.js:40:5)
```

AaronDewes:
To run Umbrel on a custom domain, change Line 57 of scripts/start to include your custom domain

grmkris:
I did that.. I changed that to:
`DEVICE_HOSTS="http://${DEVICE_IP},http://${DEVICE_HOSTNAME}.local,https://${DEVICE_HOSTNAME}.local,http://${DEVICE_HOSTNAME},https://${DEVICE_HOSTNAME}"`  

`DEVICE_HOSTS="http://${DEVICE_IP},http://${DEVICE_HOSTNAME}.local,https://${DEVICE_HOSTNAME}.local,http://${DEVICE_HOSTNAME},https://${DEVICE_HOSTNAME},https//umbrel.payserver.org/,umbrel.payserver.org"`

grmkris:
This is the ouput if I exec into docker container (umbrel manager)
[...]

PR #585, opened 2021-02-27
Add port to device hosts (for CORS)

louneskmt:
This PR adds device hosts with nginx port to the device hosts list, for CORS. Without this, setting a custom `NGINX_PORT` will lead to a "Not allowed by CORS" on dashboard login.

Fixes https://github.com/getumbrel/umbrel/issues/584, https://github.com/getumbrel/umbrel/issues/398#issuecomment-759069511.

### DIFF
--- a/scripts/start
+++ b/scripts/start
@@ -55,6 +55,7 @@ echo
 DEVICE_IP="$(ip -o route get to 8.8.8.8 | sed -n 's/.*src \([0-9.]\+\).*/\1/p')"
 DEVICE_HOSTNAME="$(hostname)"
 DEVICE_HOSTS="http://${DEVICE_IP},http://${DEVICE_HOSTNAME}.local,https://${DEVICE_HOSTNAME}.local,http://${DEVICE_HOSTNAME},https://${DEVICE_HOSTNAME}"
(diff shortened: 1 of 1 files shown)


phenomeus:
./start: line 58: NGINX_PORT: unbound variable
./start: line 58: NGINX_PORT: unbound variable

its still missing NGINX_PORT variable. if I write NGINX_PORT=8000 above its starting.

junderw:
This solution is insufficient, since NGINX_PORT is persisted in the configure script. (in the `.env` file)

So if I run on first boot with NGINX_PORT=8000, 8000 gets persisted, but if I run NGINX_PORT=8001 on second run, it gets ignored and the 8000 persisted in .env is prioritized. However, this DEVICE_HOSTS will use 8001.

louneskmt:
@junderw Hey! This a known issue, we will make a PR soon to fix it.

phenomeus:
still an issue with 0.38

louneskmt:
@phenomeus @junderw Will be fixed with #713.

mayankchhabra:
Closing this PR as the CORS check in `umbrel-manager` was updated a while ago. Thanks all!

PR #587, opened 2021-02-27
Migrating Docker filesystem to SSD.

downet:
Adding part for migration Docker filesystem to SSD.

Don't know how to correctly handle the restart of Umbrel due to restart of Docker, but.. let's see your comments.

I assume that I will also need to add the part with Docker path to umbrel-os as well.

### DIFF
--- a/scripts/update/01-run.sh
+++ b/scripts/update/01-run.sh
@@ -75,6 +75,51 @@ EOF
 # Checkout to the new release
 cd "$UMBREL_ROOT"/.umbrel-"$RELEASE"
 
+cat <<EOF > "$UMBREL_ROOT"/statuses/update-status.json
+{"state": "installing", "progress": 38, "description": "Migrating Docker filesystem to SSD", "updateTo": "$RELEASE"}
+EOF
+
+# migrating docker to SSD
+echo "Migrating Docker to SSD"
+if [[ $(grep ExecStart /lib/systemd/system/docker.service | grep -c "/mnt/data/docker") -eq 0 ]]; then
+  # update systemd service for Docker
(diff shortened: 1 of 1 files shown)


downet:
I assume there should be some check only for RPI users, since if someone is running it standalone, then he most probably dont want to move /var/lib/docker to SSD :)

AaronDewes:
Good idea and concept, however, this houlld not only happen during updates.

downet:
> Good idea and concept, however, this houlld not only happen during updates.

Feel free to advice where else I should look :)

AaronDewes:
I'd implement it in https://github.com/getumbrel/umbrel/blob/master/scripts/umbrel-os/external-storage/mount (At [this position](https://github.com/getumbrel/umbrel/blob/master/scripts/umbrel-os/external-storage/mount#L205),  which only runs on Umbrel OS anyway, but also a bit simpler:

```
systemctl stop docker
cp /var/lib/docker/ /mnt/data/docker/
rm -rf * /var/lib/docker/*
mount --bind /mnt/data/docker /var/lib/docker
systemctl start docker
```

lukechilds:
We should change the startup scripts to make sure the mount script always runs before Docker, so there's no need to stop Docker, and also that means there won't be issues with future boots with Docker starting before /var/lib/docker has been bind mounted.

Look at how we do this Umbrel root, we copy the Umbrel root from the SD card to /mnt/data/umbrel on the SSD, then bind mount the SSD dir over the SD card dir.

We should do the same for `/var/lib/docker`. Apart from instead of this happening at SSD formatting, it needs to happen on any boot if it hasn't already been setup.

PR #589, opened 2021-02-28
Move docker data to the SSD

### LINKED ISSUE #630: Move the Docker data dir to the SSD
Currently the Docker data dir lives on the SD card. This means any filesystem changes inside the container (not on volumes) are being written to the SD card not the SSD which will likely result on poor container performance and degrade SD card life.

It also means all Docker images are being stored on the SD card too which is an issue as most SD cards are quite small and some of the app Docker images are quite large.

We should move the Docker data dir to the SSD to resolve this.

We could even resolve #629 too by using one of the file based log drivers for the container logs so they end up in the Docker data dir which is now on the SSD.

AaronDewes:
Resolves #630 

Replaces #587, https://github.com/getumbrel/umbrel-os/pull/187


scripts/umbrel-os/external-storage/mount
```diff
@@ -200,6 +200,19 @@ main () {
...
+  if [[ -f "${MOUNT_POINT}/docker" ]]; then
+    echo "Yes, docker data is on the SSD"
+  else 
+    echo "No, docker data is on the SD"
+    echo "Moving data..."
+    cp  --recursive \
+        --archive \
+        --no-target-directory \
+        /var/lib/docker /mnt/data/docker
+    mount --bind /mnt/data/docker /var/lib/docker
```
louneskmt, line 213:
```suggestion
        /var/lib/docker ${MOUNT_POINT}/docker
    mount --bind ${MOUNT_POINT}/docker /var/lib/docker
```

scripts/umbrel-os/external-storage/mount
```diff
@@ -200,6 +200,19 @@ main () {
...
+  if [[ -f "${MOUNT_POINT}/docker" ]]; then
+    echo "Yes, docker data is on the SSD"
+  else 
+    echo "No, docker data is on the SD"
+    echo "Moving data..."
+    cp  --recursive \
+        --archive \
+        --no-target-directory \
+        /var/lib/docker ${MOUNT_POINT}/docker
+    mount --bind ${MOUNT_POINT}/docker /var/lib/docker
```
lukechilds, line 213: Looks like this is only ever externally mounting the directory on the initial setup and will fall back to the SD card directory on subsequent boots.

We should move the bind mount line outside the conditional checks. Only the copy is conditional, we always want to run the bind mount.

scripts/umbrel-os/external-storage/mount
```diff
@@ -200,6 +200,19 @@ main () {
   mkswap "${SWAP_FILE}"
   swapon "${SWAP_FILE}"
 
+  echo "Checking if docker containers are on the SSD"
```
lukechilds, line 203: Just to keep inline with other terminology.
```suggestion
  echo "Checking if Docker data dir is on external storage..."
```

scripts/umbrel-os/external-storage/mount
lukechilds: We should also do a quick double check the mount has succeeded. You can see how we do that above with `$UMBREL_ROOT`/`$SWAP_DIR`.

lukechilds (changes requested):

scripts/umbrel-os/external-storage/mount
```diff
@@ -200,6 +200,19 @@ main () {
...
+  if [[ -f "${MOUNT_POINT}/docker" ]]; then
+    echo "Yes, docker data is on the SSD"
+  else 
+    echo "No, docker data is on the SD"
```
lukechilds, line 207: I miss-read SD for SSD.

```suggestion
    echo "No, Docker data dir is on the SD card"
```

scripts/umbrel-os/external-storage/mount
```diff
@@ -200,6 +200,19 @@ main () {
   mkswap "${SWAP_FILE}"
   swapon "${SWAP_FILE}"
 
+  echo "Checking if docker containers are on the SSD"
+  if [[ -f "${MOUNT_POINT}/docker" ]]; then
+    echo "Yes, docker data is on the SSD"
```
lukechilds, line 205:
```suggestion
    echo "Yes, Docker data dir is on the SSD"
```

lukechilds (changes requested):

scripts/umbrel-os/external-storage/mount
AaronDewes: Done.

scripts/umbrel-os/external-storage/mount
```diff
@@ -200,6 +200,22 @@ main () {
...
+  echo "Checking if Docker data dir is on external storage..."
+  if [[ -f "${MOUNT_POINT}/docker" ]]; then
+    echo "Yes, Docker data dir is on the SSD"
+  else 
+    echo "No, Docker data dir is on the SD card"
+    echo "Moving data..."
+    cp  --recursive \
+        --archive \
+        --no-target-directory \
+        /var/lib/docker ${MOUNT_POINT}/docker
```
lukechilds, line 212:
```suggestion
        /var/lib/docker "${MOUNT_POINT}/docker"
```

scripts/umbrel-os/external-storage/mount
```diff
@@ -200,6 +200,22 @@ main () {
...
+  fi
+  mount --bind ${MOUNT_POINT}/docker /var/lib/docker
```
lukechilds, line 214: You should always quote variables to prevent string splitting in bash.

```suggestion
  mount --bind "${MOUNT_POINT}/docker" /var/lib/docker
```

For example if `MOUNT_POINT` was `/mnt/Samsung T5` then you'd end up running

```
mount --bind /mnt/Samsung T5/docker /var/lib/docker
```

instead of

```
mount --bind "/mnt/Samsung T5/docker" /var/lib/docker
```

lukechilds (changes requested):

scripts/update/01-run.sh
```diff
@@ -113,6 +113,40 @@ EOF
...
+# Move Docker data dir to external storage now if this is an old install.
+# This is only needed temporarily until all users have transitioned Docker to SSD.
+DOCKER_DIR="/var/lib/docker"
+MOUNT_POINT="/mnt/data"
+EXTERNAL_DOCKER_DIR="${MOUNT_POINT}/docker"
+if [[ ! -z "${UMBREL_OS:-}" ]] && [[ ! -d "${EXTERNAL_DOCKER_DIR}" ]]; then
+  echo "Attempting to move Docker to external storage..."
+
+  echo "Stopping Docker service..."
+  service docker stop
```
AaronDewes, line 125:
```suggestion
  systemctl stop docker
```
Not really important, but I'd use `systemctl`, while service is compatible, according to the man page, it was made for sysV, so I wouldn't use it.
Same below.

lukechilds: Good spot!

Resolved in ca8a988

lukechilds:
Note to self: We need to update the system services during OTA otherwise the bind mount isn't guaranteed to be setup before Docker is running on the next reboot.

lukechilds:
Thanks @AaronDewes!

I've refactored it a bit and added some extra logic to transition existing users over to an SSD Docker  install during OTA.

lukechilds (approved):

PR #604, opened 2021-03-02
WIP: Switch to btrfs

### LINKED ISSUE #631: Use more resilient filesystem
The Pi is pretty sensitive to power issues and can sometimes crash when used with an unstable source of power. In this event filesystem corruption is not that uncommon and can leave the device in an usable state or can take several hours for apps to recover.

Currently we use ext4 as the filesystem. We should experiment with Btrfs or ZFS and see if they are effective at reducing the chance of filesystem corruption in the event of power loss.

We should try repeatedly pulling the power with ext4/Btrfs/ZFS formatted Umbrels and seeing if there is a noticeable difference in the amount of times filesystem corruption is experienced.

AaronDewes:
Fixes #631 together with https://github.com/getumbrel/umbrel-os/pull/185. This should make Umbrel more reliable.

### DIFF
--- a/scripts/umbrel-os/external-storage/mount
+++ b/scripts/umbrel-os/external-storage/mount
@@ -59,7 +59,7 @@ get_block_device_model () {
   echo "$(echo $vendor) $(echo $model)"
 }
 
-is_partition_ext4 () {
+is_partition_ext4_or_btrfs () {
   partition_path="${1}"
   # We need to run sync here to make sure the filesystem is reflecting the
   # the latest changes in /dev/*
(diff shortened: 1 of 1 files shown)


AaronDewes:
The formatting requires btrfs-progs on Umbrel OS, which is part of https://github.com/getumbrel/umbrel-os/pull/185. Because only Umbrel OS versions released after that update will ever format to btrfs, I didn't add this to OTA updates

AaronDewes:
Just FyI: Umbrel OS v2 will allow us to choose any file system for the SD card.

PR #605, opened 2021-03-02
build: bump getumbrel/electrs hash

RandyMcMillan:
This is necessary so raspi os can pull the getumbrel/electrs image from hub.docker.com


AaronDewes:
It worked fine for me, what exactly is your issue?

AaronDewes:
Are you maybe using the armhf/32 bit version of RPi OS? That isn't supported.

PR #606, opened 2021-03-02
docker:revert docker-compose version

RandyMcMillan:
This is necessary to build/deploy Umbrel on raspi os from the git repository.


AaronDewes:
No, it isn't, you shouldn't install docker-compose via apt, but via `sudo pip3 install docker-compose`. Same with docker, use the official docker install script.

PR #627, opened 2021-03-09
Option to disable password change

jon4hz:
Optional customization are done by environment variables which aren't set by default. So the standard user shouldn't be affected.

### DIFF
--- a/scripts/start
+++ b/scripts/start
@@ -54,7 +54,7 @@ echo
 # Whitelist device IP, hostname and hidden service for CORS
 DEVICE_IP="$(ip -o route get to 8.8.8.8 | sed -n 's/.*src \([0-9.]\+\).*/\1/p')"
 DEVICE_HOSTNAME="$(hostname)"
-DEVICE_HOSTS="http://${DEVICE_IP},http://${DEVICE_HOSTNAME}.local,https://${DEVICE_HOSTNAME}.local,http://${DEVICE_HOSTNAME},https://${DEVICE_HOSTNAME},${ADDITIONAL_UMBREL_HOSTS}"
+DEVICE_HOSTS="http://${DEVICE_IP},http://${DEVICE_HOSTNAME}.local,https://${DEVICE_HOSTNAME}.local,http://${DEVICE_HOSTNAME},https://${DEVICE_HOSTNAME}"
 if [[ -f "${UMBREL_ROOT}/tor/data/web/hostname" ]]; then
     hidden_service_url=$(cat "${UMBREL_ROOT}/tor/data/web/hostname")
     DEVICE_HOSTS="${DEVICE_HOSTS},http://${hidden_service_url}"
--- a/scripts/start
+++ b/scripts/start
@@ -54,7 +54,7 @@ echo
 # Whitelist device IP, hostname and hidden service for CORS
 DEVICE_IP="$(ip -o route get to 8.8.8.8 | sed -n 's/.*src \([0-9.]\+\).*/\1/p')"
 DEVICE_HOSTNAME="$(hostname)"
-DEVICE_HOSTS="http://${DEVICE_IP},http://${DEVICE_HOSTNAME}.local,https://${DEVICE_HOSTNAME}.local,http://${DEVICE_HOSTNAME},https://${DEVICE_HOSTNAME}"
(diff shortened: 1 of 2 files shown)


louneskmt:
Why disable password change?

scripts/start
```diff
@@ -54,7 +54,7 @@ echo
 # Whitelist device IP, hostname and hidden service for CORS
 DEVICE_IP="$(ip -o route get to 8.8.8.8 | sed -n 's/.*src \([0-9.]\+\).*/\1/p')"
 DEVICE_HOSTNAME="$(hostname)"
-DEVICE_HOSTS="http://${DEVICE_IP},http://${DEVICE_HOSTNAME}.local,https://${DEVICE_HOSTNAME}.local,http://${DEVICE_HOSTNAME},https://${DEVICE_HOSTNAME}"
+DEVICE_HOSTS="http://${DEVICE_IP},http://${DEVICE_HOSTNAME}.local,https://${DEVICE_HOSTNAME}.local,http://${DEVICE_HOSTNAME},https://${DEVICE_HOSTNAME},${ADDITIONAL_UMBREL_HOSTS}"
```
AaronDewes, line 57: Umbrel PRs should only contain one small change per PR, so this should be separate.

junderw:
NACK password change disable

ACK additional hosts

Why is this one PR?

louneskmt:
Same as @junderw, NACK password change disable. 

Btw, to add a new external Umbrel env var, you need to add it to the `scripts/configure` file, `scripts/start` (when it calls the configure script), `templates/.env.template` file. Check other variables like NETWORK to see how it is handled.

jon4hz:
> Why disable password change?

IMO there should be an option to disable the password overwrite for several reasons. 

- People should be able to choose a different password for SSH since that is a completely different service than Umbrel itself. 
- By using the different passwords for the web-based application and for the login to the host itself the security can be enhanced.
- Generally is Umbrel "just" a service and changing such important settings should  be possible without having them overwritten on every update.

In addition I think that change wont bring any disadvantage. By default the dashboard password will be used as SSH-password. A user which disables that option deliberately will know what he's doing.

junderw:
I still don't agree this is a good idea. I see no benefit in the project's current state.

- Right now the password is used for two things, the umbrel user password and lnd wallet password.
- All of your stated goals should be accomplished with more nuanced (and more complicated) changes.
- If the logic "user which disables that option deliberately will know what he's doing" stands, then we can say "that user also knows how to ssh in, add an ssh pubkey to login, and disable ssh password auth... so no changes are needed.
- This change would need to be persisted. See previous comments.
- This change would need to be reflected in the UI. With this current change as is, users can change the password and it will say "success" but the underlying process that actually changes the password will not run, causing the user to potentially delete old password from password manager etc. and leading to fund loss.

stevenroose:
@jon4hz just a heads up that if you don't have explicit individual approval by all the contributors to this repository to publish your modified source code ("Adapted Material"), you are officially in violation of the license of this project and you can be sued:

AaronDewes:
@stevenroose Just a heads up that you are annoying.

AaronDewes:
@stevenroose Just another head up that you should learn to read: https://github.com/getumbrel/umbrel/issues/291#issuecomment-711153942

stevenroose:
> @stevenroose Just another head up that you should learn to read: [#291 (comment)](https://github.com/getumbrel/umbrel/issues/291#issuecomment-711153942)

Well, he is not the only copyright holder of the code inside the repository. Every contributor that contributed code to the repository has copyright over its contribution published under the license.

> @stevenroose Just a heads up that you are annoying.

Sure, I realize that. I just think this kind of precedent is something we really cannot accept in our ecosystem. Not in FOSS and certainly not in Bitcoin. The FOSS community holds grudges towards some projects using the Apache 2 or GPL license, while they are still far less restrictive as the license used by this project.

stevenroose:
Don't get me wrong, I personally disagree with the entire notion of copyright and software licenses in the first place. But the fact that a Bitcoin project like Umbrel would use such a system that's based on systematic violence to effectively discourage individuals from contributing to their software is quite striking.

jon4hz:
> @stevenroose Just a heads up that you are annoying.

Very mature answer. 

@stevenroose Thanks for the heads up. Right now I'm considering switching to mynode anyway. I don't like Umbrel's philosophy, development or licensing. 

@devs you can close the PR because the majority is against it.

jon4hz:
Closing this as my contribution is basically illegal under this trash license

PR #634, opened 2021-03-12
Retry Umbrel systemd startup service indefinitely

### LINKED ISSUE #633: Network may be temporarily unreachable after boot
```
RTNETLINK answers: Network is unreachable
```

This message appears in the logs of a few users, so it seems like sometimes, even though we normally wait for the network to be online. This can probably be fixed by retrying the startup a few times.

AaronDewes:
Fixes #633


lukechilds:
Awesome stuff @AaronDewes, tested and working great.

I removed the max 5 retries and time limit so it just attempts to restart indefinitely until it succeeds. It's possible a user could power on the device, and not plug the network cable in until a day later so I think it makes sense to just keep retrying until networking is up.

Also reduced the interval from 30 seconds to 10 seconds so in the event the user attaches power and then networking but missed the first startup attempt, they aren't left waiting too long.

Let me know if you think these changes are ok or if I misunderstood something!

lukechilds (approved):

PR #638, opened 2021-03-13
Update systemd services on updates

AaronDewes:
Currently if we update services in the getumbrel/umbrel repo, they will get installed on new devices, but not updated on old devices. This PR updates all systemd services during OTA update, and installs any new ones.


lukechilds:
Thanks @AaronDewes!

PR #640, opened 2021-03-13
Disable referrer

### LINKED ISSUE #639: Clicking on an external link discloses onion address to third-party
When browsing to Umbrel on tor, several hyperlinks refer to  external websites (e.g. getumbrel.com, github.com and t.me). As no Referer policy is configured, the onion address of an umbrel installation is disclosed to the external website through the Referer header. Also no other security headers are present. 

I suggest to include: Referrer-Policy: no-referrer in all responses and configure other security headers.

AaronDewes:
Fixes #639

### DIFF
--- a/templates/nginx-sample.conf
+++ b/templates/nginx-sample.conf
@@ -37,6 +37,7 @@ http {
 			proxy_http_version 1.1;
 			proxy_set_header Upgrade $http_upgrade;
 			proxy_set_header Connection "upgrade";
+			proxy_set_header Referrer-Policy "no-referrer";
 		}
 	}
 }


templates/nginx-sample.conf
```diff
@@ -37,6 +37,7 @@ http {
 			proxy_http_version 1.1;
 			proxy_set_header Upgrade $http_upgrade;
 			proxy_set_header Connection "upgrade";
+			proxy_set_header Referrer-Policy "no-referrer";
```
lukechilds, line 40: This is setting the header on the proxied request to the dashboard web server, not the response to the browser.

I think this should be handled in the dashboard container since it's directly related to that.

lukechilds (changes requested):

PR #646, opened 2021-03-16
Update LICENSE.md

elkimek:
https://twitter.com/mynodebtc/status/1371872750390894602?s=20

Adding the SECTION 9!!!1337!!!


AaronDewes:
If you have a license change suggestion, please just tell us. If you just open a PR and change the license, that is not legally allowed under the license, so please do not do it.

PR #664, opened 2021-03-25
Revert "Update Tor to 0.4.5.7"

lukechilds:
This reverts commit d86e209d5b24f9403b8e13ed2276b9626f2f46a9.

There appears to be a bug in the amd64 Docker image.

// cc @nolim1t

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,7 +7,7 @@ x-logging: &default-logging
 services:
         tor:
                 container_name: tor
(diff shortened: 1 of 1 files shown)


AaronDewes:
Can you try 0.4.4.8? Maybe that one works?

@nolim1t The amd64 image seems to contain arm64 binaries.

nolim1t:
the hash seems different though..  and I'm using docker buildx.

Maybe I'll try building again

lukechilds:
Sorry @nolim1t, ignore that, the amd64 image hash was pinned in this PR instead of the multiarch build manifest hash!

lukechilds:
Resolved with https://github.com/getumbrel/umbrel/pull/665

PR #677, opened 2021-03-26
Update loop service to v0.12.1-beta in ride-the-lightning app

louneskmt:


lukechilds:
Maybe we should transition to https://hub.docker.com/r/lightninglabs/loopserver as advertised in https://github.com/lightninglabs/loop/releases/tag/v0.12.1-beta.

Although it's odd they don't appear use the same version numbers 🤔

louneskmt:
@lukechilds Will look into it, thanks!

Issue #687, opened 2021-03-29
Not allowed by CORS


SmartCorporate:
This alarm come out either when I finish entering the 24 words or even if I skip to them later, I looked at the old issue but they didn't solve my problem, can you help me? I am using the latest version of UBUNTU

louneskmt:
What url are you accessing from?

phenomeus:
https://github.com/getumbrel/umbrel/issues/398 ?

SmartCorporate:
@phenomeus , thx #398 , but not solving. Any other idea?

phenomeus:
not anymore, I have this issue since 0.34 or something, no I even can't fix it with mentioned workaround in 0.38

ChowSinWon:
didn't try the actual fix #708 but as far as i could see the idea is to add the port that has been set with NGINX_PORT to the allowed hosts (like in #398).
i added the port by myself and it worked for me.

phenomeus:
where did you added this? @ChowSinWon

ChowSinWon:
@phenomeus i added `:${NGINX_PORT}` to every host in the DEVICE_HOSTS-definition in `scripts/start`
like that:
```
DEVICE_HOSTS="http://${DEVICE_IP}:${NGINX_PORT},http://${DEVICE_HOSTNAME}.local:${NGINX_PORT},https://${DEVICE_HOSTNAME}.local:${NGINX_PORT},http://${DEVICE_HOSTNAME}:${NGINX_PORT},https://${DEVICE_HOSTNAME}:${NGINX_PORT}"
```

because i have on that system a service that is already listening on port 80 i need to set a different port when starting up umbrel:
```
sudo NGINX_PORT=3080 ./scripts/start
```

btw. the onion/tor address does not need that port

phenomeus:
*runningtotrythisout*

phenomeus:
workaround is not working anymore with 0.4.2 @ChowSinWon @lukechilds @louneskmt 

any chance to fix this once and for all?

DRiOscar:
Hi @phenomeus,

Mostly update Umbrel by CLI mode
```
sudo ./scripts/stop
sudo update/update --repo getumbrel/umbrel#v0.4.2
# modify the scripts/start with the PR diff
sudo NGINX_PORT=$YOUR_NGINX_PORT ./scripts/start
```

nmfretz:
closing old support issue

PR #708, opened 2021-04-12
Fix for Not allowed by CORS

xlf42:
Change fixes the "Not allowed by CORS" issue described in https://github.com/getumbrel/umbrel/issues/687 for me.
Would appreciate feedback 
- whether the fix works for others
- whether there is a better solution
Thanks
Alex


matthew-morrison:
I tried this on the current release of umbrel v0.3.9 on a fresh download of umbrel after getting this CORS error, and received the following errors:

`scripts/start: line 59: NGINX_PORT: unbound variable
scripts/start: line 59: NGINX_PORT: unbound variable`.

xlf42:
Hi,

I looked into how NGINX_PORT is supposed to be set and found.
At the end, I decided to add a snipped for NGINX_PORT and NETWORK directly.

HTH

> I tried this on the current release of umbrel v0.3.9 on a fresh download of umbrel after getting this CORS error, and received the following errors:
> 
> `scripts/start: line 59: NGINX_PORT: unbound variable scripts/start: line 59: NGINX_PORT: unbound variable`.

phenomeus:
@team why is it taking so long to fix the custom nginx situation? :/

scripts/start
```diff
@@ -41,9 +41,13 @@ if [[ ! -d "$UMBREL_ROOT" ]]; then
...
 fi
 
+# make sure NGINX_PORT andd NETWORK are set
+NGINX_PORT="${NGINX_PORT:-80}"
+NETWORK="${NETWORK:-mainnet}"
+
 # Configure Umbrel if it isn't already configured
 if [[ ! -f "${UMBREL_ROOT}/statuses/configured" ]]; then
-  NGINX_PORT=${NGINX_PORT:-80} NETWORK="${NETWORK:-mainnet}" "${UMBREL_ROOT}/scripts/configure"
+  "${UMBREL_ROOT}/scripts/configure"
```
AaronDewes, line 50: Please keep the code changes for this PR as low as possible because of Umbels 1 PR = 1 Feature/Bug Fix policy, this doesn't need to be changed to fix the issue.

AaronDewes (changes requested):

dugwood:
@xlf42 I think the issue here is to use the NGINX_PORT variable, as it may be used only the first time. All other calls shouldn't use NGINX_PORT at all.

Hence, a solution would be a line just after this test:
https://github.com/getumbrel/umbrel/blob/b2c034c95ed7cbcdc378955cdd0eee63a74eca3d/scripts/start#L45

such as:

    NGINX_PORT=`grep "NGINX_PORT" "${UMBREL_ROOT}/.env"|sed 's/.*=//'`

which would read the real value from the `.env` file.

Else your code is great :-)

scripts/start
```diff
@@ -59,13 +59,15 @@ echo
 # Whitelist device IP, hostname and hidden service for CORS
 DEVICE_IP="$(ip -o route get to 8.8.8.8 | sed -n 's/.*src \([0-9.]\+\).*/\1/p')"
 DEVICE_HOSTNAME="$(hostname)"
+DEVICE_DOMAIN=`host  "${DEVICE_HOSTNAME}" | awk '{print $1}' | sed 's/^[^.]*\.//' | head -1`
```
19wolf, line 62: I think I've narrowed down my issue to this line. On my machine, the script stops at this line. My hostname is `nephele` so `host $(hostname)` returns an error.

19wolf: After changing this line to `DEVICE_DOMAIN=local` I'm still having issues, regardless of if I connect to the IP or domain.

xlf42:
Is this PR still helpful or did some other PR solve it?

phenomeus:
Still not fixed

mayankchhabra:
Sorry for the unneccesary long delay in responding, folks! Thanks for working on it @xlf42. The CORS check was updated in umbrel-manager a while ago, so hopefully this issue should be gone for good. I'm closing this PR for now, but if anyone is still facing an issue please let me know.

PR #709, opened 2021-04-12
Disable journald logging

### LINKED ISSUE #629: Write logs to SSD
Currently we write all application logs to syslog which lives on the SD card.

Syslog is probably not the best place for this. It's fine having syslog on the SD card with minimal OS log writes, but all the heavy Umbrel specific log stuff should be written to the SSD.

Probable as a single or multiple file inside `$UMBREL_ROOT/logs` which already lives on the SSD. Then we can just `tail -f $UMBREL_ROOT/logs/*` to watch all Umbrel logs.

lukechilds:
Resolves #629 

We were writing all container logs to journald which in turn writes to syslog which means all logs ultimately get written to the SD card on Umbrel OS. Some contianers make pretty heavy use of logging. This is bad for performance and also reduces the life and reliability of the SD card.

By disabling journald we fall back to the default json-file based Docker logging driver which stores logs in JSON files in the Docker data directory. Since https://github.com/getumbrel/umbrel/pull/589 the Docker data directory now lives on the external storage device on Umbrel OS which means perf will be much better and won't degrade SD card lifetime.

### DIFF
--- a/apps/README.md
+++ b/apps/README.md
@@ -96,16 +96,9 @@ Let's copy-paste the following template `docker-compose.yml` file in a text edit
 ```yml
 version: "3.7"
 
-x-logging:
-  &default-logging
-  driver: journald
-  options:
-    tag: "umbrel-app {{.Name}}"
-
 services:
   web:
     image: <docker-image>:<tag>
-    logging: *default-logging
     restart: on-failure
     stop_grace_period: 1m
     ports:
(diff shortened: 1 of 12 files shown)


AaronDewes (approved):
Just a reminder that these PRs need to be updated too:
https://github.com/getumbrel/umbrel/pull/385
https://github.com/getumbrel/umbrel/pull/625
https://github.com/getumbrel/umbrel/pull/566 (I'll do this one)

lukechilds:
Good spot, thanks @AaronDewes.

PR #391, opened 2021-01-02
Implement debug execution from docker containers 

louneskmt:
This PR adds a trigger to execute the debug script and save the output.

Related PRs:
- https://github.com/getumbrel/umbrel-dashboard/pull/275 (main)
- https://github.com/getumbrel/umbrel-manager/pull/66

PR #718, opened 2021-04-14
Remove web logs

lukechilds:
Related: https://github.com/getumbrel/umbrel-dashboard/pull/330

This just tails syslog, which doesn't contain much useful information any more since https://github.com/getumbrel/umbrel/pull/709.

Instead, if users want to view the logs, they should view debug information via https://github.com/getumbrel/umbrel/pull/391.


AaronDewes:
Okay, I'll close my UI improvements PR in umbrel-frontail then.

AaronDewes:
Don't forget to remove this from SECURITY.md too

lukechilds:
>Okay, I'll close my UI improvements PR in umbrel-frontail then.

Ah yes sorry about that. It would be quite a bit of work to continue using frontail after #709. And it doesn't make sense to work on it when #391 more or less achieves the same without the data leak issue.

>Don't forget to remove this from SECURITY.md too

Thanks, completely forget about that!

AaronDewes:
Yes, #391 is great. @mayankchhabra would be great to get a design review / design improvements from you for the dashboard debug UI.

@louneskmt And you can probably remove the debug option from the loading screen, the loading screen was nearly removed.

PR #741, opened 2021-04-24
Ensure SD card updater runs after networking and Docker

AaronDewes:
The external storage update requires entwork & docker to be running, if it runs before they are ready it will fail to update the SSD installation if the SD card installation is newer.


lukechilds:
This is working great @AaronDewes, thanks!

Some minor tweaks:
- It's nice to keep the explicit dep for all services: https://github.com/getumbrel/umbrel/pull/741/commits/02cbfdeb4ea1030be42b63457f5bdc3c7aacea18
- Renamed to include SD card, just incase it's confusing to people not familair with the codebase, it might seem like it's some sort of update monitoring process: https://github.com/getumbrel/umbrel/pull/741/commits/e540f9738519ee56005e020fb2901785c2d37aaa

PR #749, opened 2021-04-26
Apply strict formatting for all compose files

louneskmt:
This PR applies a strict and fix formatting for all Umbrel and apps compose files. This was necessary as the previous formatting was very inconsistent and different across all files.

It needs careful reviews and testing to check everything still works and hasn't been impacted by one of these changes.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,4 @@
-version: '3.7'
+version: "3.7"
 
 services:
     tor:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,188 +1,197 @@
 version: '3.7'
 
 services:
-        tor:
-                container_name: tor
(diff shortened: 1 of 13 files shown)


docker-compose.yml
```diff
@@ -1,188 +1,197 @@
 version: '3.7'
```
AaronDewes, line 1:
```suggestion
version: "3.7"
```

This is using " in every app, so we should use it here too.

AaronDewes (approved):
Only one small change.

docker-compose.yml
louneskmt: Sure, I missed this one 😁

AaronDewes (approved):

louneskmt:
@AaronDewes Have you been able to test this? Trying now.

AaronDewes:
It started fine for me on a custom Ubuntu setup, but I haven't done much testing yet.

louneskmt:
Ideally, we should try to install each app and test them all, to see if these changes didn't mess up with some app configuration.

PR #752, opened 2021-04-28
Allow running without root permissions

AaronDewes:
This still needs more testing, but now the start script can be run without root, so only docker permissions are required.

---

I forgot that some scripts like the backup monitor require root.

### DIFF
--- a/karen
+++ b/karen
@@ -3,9 +3,9 @@
 # karen watches for signals and executes triggers in the events dir
 # karen gets triggered a lot
 
-check_root () {
-  if [[ $UID != 0 ]]; then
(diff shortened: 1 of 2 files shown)


lukechilds:
Shutdown and restart also require root.

AaronDewes:
> 
> 
> Shutdown and restart also require root.

`systemctl poweroff` and `systemctl reboot` should work without root, but we would loose compatibility with non-systemd systems then.

HenryWait:
podman is a replacement for docker that can run rootless, consideration to switch to podman should have some thought.

PR #753, opened 2021-04-28
Use same tor for hashed password as we use later

AaronDewes:
This avoids nodes having to download Tor when we it's bundled with Umbrel OS


AaronDewes:
@lukechilds This is a small bugfix which makes the entwork only required at a later stage of boot and makes starting Umbrel a bit faster. Can we get this into 0.3.10?

PR #756, opened 2021-04-28
Don't force usb-storage driver on incompatible devices

lukechilds:
Since https://github.com/getumbrel/umbrel/pull/722 we now disable the UAS driver for all storage devices and instead use the older usb-storage driver. It seems that some users running non-standard are running into issues where they have storage devices that do not work with the usb-storage driver and are unable to mount their drive.

This PR defaults to the usb-storage driver, but if the device fails to be enumerated, falls back to the UAS driver.


lukechilds:
Closing as this doesn't appear to fix the issue: https://github.com/getumbrel/umbrel/commit/3399ec52efe9f605a8e9756b2e0623efe702b364#commitcomment-50133173

PR #789, opened 2021-05-11
Allow usage of NVME drives as external storage for BitPiRat

AaronDewes:
This is required for the BitPiRat and useful for the MirkoPC to use NVME drives for Umbrel.

---

My SSD for testing will arrive next week so I can properly test this.


AaronDewes:
Are you on telegram or Twitter? I built something for this, but didn't get my CM4 to work so I could test it. If you test ity we can get this into Umbrel.

AaronDewes:
@lukechilds @mayankchhabra This PR is now confirmed to work and ready for review.

PR #791, opened 2021-05-12
Bump thunderhub version

apotdevin:


AaronDewes:
Thanks a lot! @lukechilds we should get this into the next release.

@apotdevin We'll include such updates before making a release ourselves in the future to avoid you as a maintainer having to update our code every time you make a release. We'll soon have a bot which takes care of the update check for us, so we can just quickly update all packages. Here's an example of how it will work (Not enabled in this repo yet): https://github.com/UmbrelBuilds/umbrel-manager/pull/3#issuecomment-846543913

lukechilds:
Thanks for this @apotdevin!

Apologies for the delay, we've been focusing on some stability features but we'll get this tested and merged for the next release!

lukechilds (approved):
Thanks @apotdevin, sorry for the delay!

PR #797, opened 2021-05-16
Update the debug script

AaronDewes:
I still don't like what I did in the original debug script, so here's version 3.

### Update for debug server v3
The new server is in development. It seperates the frontend from the backend.
Both parts of the server have been completely rewritten to look better. MongoDB Atlas is used as the backend for the server v3 instead of Firebase.
The legacy debug.umbrel.tech will be deleted a week after 0.3.11 is released (The domain will be removed in a few minutes, but umbrel-paste.vercel.app is the only server we use anyway)

### Remove raspinfo copyright
While the script was based on Raspinfo in the beginning, none of the raspinfo code is left inside our codebase anymore
### Remove duplicated result if multiple SSDs were connected

```
================
==== Result ====
================
```

This was shown twice before.

### Remove firmware version from the output

This may be useful to RPi Foundation staff (which is why it was in raspinfo) but probably nothing we'll need.

### DIFF
--- a/events/triggers/debug
+++ b/events/triggers/debug
@@ -11,7 +11,7 @@ json_encode () {
   jq --null-input --raw-input 'reduce inputs as $line (""; . += "\($line)\n")'
 }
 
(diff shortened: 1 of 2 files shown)


louneskmt:
Why don't we upload via Tor by default?

AaronDewes:
I'll change that

PR #798, opened 2021-05-17
Use the latest build of bitcoind

AaronDewes:
No changes were made, but the dource code has been verified in a more secure way.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,7 +30,7 @@ services:
                         ipv4_address: $NGINX_IP
         bitcoin:
                 container_name: bitcoin
-                image: lncm/bitcoind:v0.21.1@sha256:5bedb46d698de16c59e9e43d31485d0d82239bd437d62ac7cf47ebb633214f37
+                image: lncm/bitcoind:v0.21.1@sha256:07fe4abfdaf752abbf943035b05538f277e4433b9a39c9e851e40fe9c0a7ea79
                 depends_on: [ tor, manager, nginx ]
                 volumes:
                         - ${PWD}/bitcoin:/data/.bitcoin


nolim1t (approved):

lukechilds:
Aren't the LNCM builds supposed to be deterministic? The hash shouldn't have changed on the image if it was built from the same source but just verified differently.

nolim1t:
Yes it is but this uses the old way of fetching and verifying both sources from different sites.

We had to rely on workarounds of using the same site.

lukechilds:
But this resulting image has a different hash?

If the verification method has changed but the resulting image is the same the hash should not have changed.

nolim1t:
According to docker its a new build irespective of changes. The way the docker images works if you run the release scripts it will be a new build number and it will get reassigned. This is so you can still refer to the old ones.

lukechilds:
The Docker images don't include tags, if the hash has changed the image is different, so either the builds aren't deterministic, or someone is potentially trying to sneak something in.

lukechilds:
The builds are not deterministic, the `bitcoind` user entry in `/etc/shadow` changes between builds.

```
$ container-diff diff daemon://bitcoind:old daemon://bitcoind:new --type=file

-----File-----

These entries have been added to bitcoind:old: None

These entries have been deleted from bitcoind:old: None

These entries have been changed between bitcoind:old and bitcoind:new:
FILE               SIZE1        SIZE2
/etc/shadow        452B         452B


$ container-diff diff daemon://bitcoind:old daemon://bitcoind:new --type=file --filename=/etc/shadow

-----File-----

These entries have been added to bitcoind:old: None

These entries have been deleted from bitcoind:old: None

These entries have been changed between bitcoind:old and bitcoind:new:
FILE               SIZE1        SIZE2
/etc/shadow        452B         452B


-----Diff of /etc/shadow-----


--- bitcoind:old
+++ bitcoind:new
@@ -28 +28 @@
-bitcoind:!:18749:0:99999:7:::
+bitcoind:!:18764:0:99999:7:::
```

lukechilds:
I'll close this since there's no need to push this out to users. It'll mean everyone will have to download a new bitcoind image during OTA that's identical to the one they already have apart from the `/etc/shadow` entry.

PR #812, opened 2021-06-05
Add status server

### LINKED ISSUE #632: Show startup errors via web UI
Currently there are many different types of errors that can occur at the OS level outside of Umbrel that the user gets no feedback on.

e.g if a SSD isn't attached, Umbrel won't start, and the user has no feedback. umbrel.local is down,. They need to SSH in and view logs to see what's wrong. We should have a webserver that displays errors via umbrel.local in cases like this.

To do this we should probably run a simple web server on the host that proxies through to the Umbrel web server. This can be running as a high priority startup service. It should be up and running as soon as possible, even before Docker or external storage mounting. I anything goes wrong with the startup process the issue should be reflected here.

This also means we can more easily show progress updates during things like OTA updates and shutdown where we want to continue to display the progress of events after the Umbrel web server has stopped.

lukechilds:
Closes #632 

This PR adds a new status server service to Umbrel OS. It runs as one of the first services at boot and will respond to all traffic on port 80. It'll show the progress of the current boot.

If an error is encountered the user is given a description of what went wrong and is able to shutdown or restart from the web UI.

The iptables rules mean that any Docker container listening on port 80 will take priority over the status server, however as soon as that Docker container is killed port 80 will fall back to the status server again.

To simulate some bootup events for testing/styling the UI you can run this one liner:

```
sudo bash -c 'while true; do echo "" > /umbrel-status; sleep 2; echo "mount:started" >> /umbrel-status; sleep 2; echo "mount:completed" >> /umbrel-status; sleep 2; echo "sdcard-update:started" >> /umbrel-status; sleep 2; echo "sdcard-update:errored:no-block-device" >> /umbrel-status; sleep 4; done'
```

You can always view the status server at port 8000, regardless of if Umbrel has started or not. When running the above script you should see this:

status-server


scripts/umbrel-os/status-server/static/index.html
```diff
@@ -0,0 +1,34 @@
...
+  <meta name="robots" content="noindex, nofollow" />
+  <meta name="referrer" content="no-referrer" />
+  <link rel="icon" href="/favicon.png">
+  <link rel="stylesheet" href="styles.css">
+  <meta name="description" content="Welcome back">
+</head>
+
+<body>
+  <noscript>
+    <strong>We're sorry but Umbrel
```
AaronDewes, line 18:
```suggestion
    <strong>We're sorry, but Umbrel
```

Isn't that required here?

scripts/umbrel-os/services/umbrel-status-server-iptables-update.service
```diff
@@ -0,0 +1,25 @@
+# Umbrel Status Server iptables Update
+# Installed at /etc/systemd/system/umbrel-status-server-iptables-update.service
+
+# This is needed because when Docker starts it apends it's own iptables rules
```
louneskmt, line 4:
```suggestion
# This is needed because when Docker starts it appends it's own iptables rules
```

scripts/umbrel-os/status-server/static/index.html
```diff
@@ -0,0 +1,34 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <title>Umbrel</title>
+  <meta charset="utf-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
```
AaronDewes, line 7: Are we still compatible with this legacy browser?

AaronDewes (dismissed):

scripts/umbrel-os/services/umbrel-status-server-iptables-update.service
```diff
@@ -0,0 +1,25 @@
...
+After=docker.service
+
+[Service]
+Type=oneshot
+ExecStart=/home/umbrel/umbrel/scripts/umbrel-os/status-server/setup-iptables
+User=root
+Group=root
+StandardOutput=syslog
+StandardError=syslog
+SyslogIdentifier=status server
```
AaronDewes, line 22:
```suggestion
SyslogIdentifier=status server iptables
```

Maybe we can use a different identifier for this?

scripts/umbrel-os/status-server/util/status_file.py
```diff
@@ -0,0 +1,54 @@
...
+# bar:started
+# fizz:started
+# fizz:completed
+# bar:errored:no-network-access
+
+# Creates an empty status file
+def create_empty(status_file_path):
+    Path(status_file_path).open('w').close()
+
+# Retuns the index of the dict with a matching ID from a list
```
louneskmt, line 18:
```suggestion
# Returns the index of the dict with a matching ID from a list
```

louneskmt (commented):
Great work on this @lukechilds!

scripts/umbrel-os/status-server/static/index.html
lukechilds: It's just pulled straight from https://github.com/getumbrel/umbrel-dashboard/blob/c36af7cb58bec8d0533178675886a63f4a1fcc88/public/index.html

scripts/umbrel-os/services/umbrel-status-server-iptables-update.service
lukechilds: Great spot!

scripts/umbrel-os/services/umbrel-status-server-iptables-update.service
lukechilds: Thanks!

scripts/umbrel-os/status-server/static/index.html
lukechilds: It's just pulled straight from https://github.com/getumbrel/umbrel-dashboard/blob/c36af7cb58bec8d0533178675886a63f4a1fcc88/public/index.html

lukechilds:
Added in the statuses for all the different services and now auto-reloading on successful boot:

https://user-images.githubusercontent.com/2123375/120927520-45876180-c70b-11eb-9be1-2f2c27f79acf.mov

scripts/umbrel-os/status-server/static/script.js
```diff
@@ -0,0 +1,151 @@
...
+    };
+    node.querySelector('.icon').classList[service.status == 'started' ? 'add' : 'remove']('rotate');
+    node.querySelector('.icon').innerText = icons[service.status];
+
+    // Render Error
+    if(service.status === 'errored' && !node.querySelector('.error')) {
+      const errorTemplates = {
+        'monitor-check': `
+          <b>Error:</b> External storage device check failed.
+          <p>This can sometimes be cause by using an unnoficial power supply.</p>
```
louneskmt, line 94:
```suggestion
          <p>This can sometimes be caused by using an unnoficial power supply.</p>
```

scripts/umbrel-os/status-server/static/script.js
```diff
@@ -0,0 +1,151 @@
...
+    };
+    node.querySelector('.icon').classList[service.status == 'started' ? 'add' : 'remove']('rotate');
+    node.querySelector('.icon').innerText = icons[service.status];
+
+    // Render Error
+    if(service.status === 'errored' && !node.querySelector('.error')) {
+      const errorTemplates = {
+        'monitor-check': `
+          <b>Error:</b> External storage device check failed.
+          <p>This can sometimes be caused by using an unnoficial power supply.</p>
```
louneskmt, line 94:
```suggestion
          <p>This can sometimes be caused by using an unofficial power supply.</p>
```

scripts/umbrel-os/status-server/static/script.js
```diff
@@ -0,0 +1,151 @@
...
+          <b>Error:</b> Umbrel couldn't find a storage device attached.
+          Please make sure your device is attached and then reboot.
+        `,
+        'multiple-block-devices': `
+          <b>Error:</b> Umbrel found multiple storage devices attached.
+          Please remove one of them and reboot.
+        `,
+        'rebinding-failed': `
+          <b>Error:</b> Umbrel failed to use your storage device in low power mode.
+          Are you using the reccomended hardware?
```
louneskmt, line 109:
```suggestion
          Are you using the recommended hardware?
```

scripts/umbrel-os/status-server/static/script.js
```diff
@@ -0,0 +1,151 @@
...
+        node.appendChild(createNode(`
+          <div class="error card fade-in">
+            ${errorTemplates[service.error]}
+          </div>
+        `));
+      }
+    }
+  });
+
+  // Togle button visiblity
```
louneskmt, line 123:
```suggestion
  // Toggle button visibility
```

scripts/umbrel-os/services/umbrel-status-server-iptables-update.service
```diff
@@ -0,0 +1,25 @@
+# Umbrel Status Server iptables Update
+# Installed at /etc/systemd/system/umbrel-status-server-iptables-update.service
+
+# This is needed because when Docker starts it appends it's own iptables rules
```
louneskmt, line 4:
```suggestion
# This is needed because when Docker starts it appends its own iptables rules
```

scripts/umbrel-os/services/umbrel-status-server-iptables-update.service
```diff
@@ -0,0 +1,25 @@
...
+# after ours. This means traffic on port 80 will never arrive at a Docker
```
louneskmt, line 5:
```suggestion
# after ours. This means traffic on port 80 will never arrive at a Docker container
```

louneskmt (commented):
Some other typos (sorry @lukechilds 😅🤣)

AaronDewes:
Please link #632 in the PR description ("Closes #632") so github can close the issue automatically

lukechilds:
Implemented shutdown/reboot UI feedback.

You can see the flow here with my booting my Umbrel without an SSD attached, then seeing the error, plugging an SSD in, then clicking the reboot button.


https://user-images.githubusercontent.com/2123375/121197660-af367580-c89b-11eb-879a-1c539aa4f8b6.mov

lukechilds:
Improved UI by the chad himself @mayankchhabra 

https://user-images.githubusercontent.com/2123375/121369309-eec99400-c965-11eb-8282-321cc2ed061f.mov

scripts/umbrel-os/status-server/static/index.html
```diff
@@ -67,28 +67,28 @@
       </svg>
 
       <div class="error-text monitor-check">
-        <p><b class="text-dark">Error: External drive not connected</b></p>
-        <p>This can sometimes be caused by using an unofficial power supply.</p>
+        <p><b class="text-dark">Error: External drive disconnected</b></p>
+        <p>The external drive was disconnected while Umbrel was running. This can sometimes happen when using an unofficial Raspberry Pi power supply.</p>
```
lukechilds, line 71: Why the Raspberry Pi distinction here? It could happen on many SBCs.

scripts/umbrel-os/status-server/static/index.html
```diff
@@ -67,28 +67,28 @@
...
       </div>
 
       <div class="error-text semver-mismatch">
-        <p><b class="text-dark">Error: Can't upgrade from SDcard due to version mismatch</b></p>
-        <p>Can't upgrade from SDcard due to version mismatch.</p>
+        <p><b class="text-dark">Error: Version mismatch</b></p>
+        <p>The version of UmbrelOS on your microSD card is different than the version of Umbrel on your external drive.</p>
```
lukechilds, line 76: Maybe we can clarify that the reason is that the installed version is too old to be directly upgraded? The error sounds like the issue is just that the versions are different.

```
The version of Umbrel OS on your microSD card is too recent to safely upgrade the version of Umbrel installed on your external drive.
```

scripts/umbrel-os/status-server/static/index.html
```diff
@@ -67,28 +67,28 @@
...
-        <p><b class="text-dark">Error: Umbrel found multiple storage devices attached</b></p>
-        <p>Please remove one of them and reboot.</p>
+        <p><b class="text-dark">Error: Multiple external drives found</b></p>
+        <p>Umbrel only works with one external drive of atleast 1TB size. Please disconnect any additional external drives and restart your Umbrel.</p>
       </div>
 
       <div class="error-text rebinding-failed">
-        <p><b class="text-dark">Error: Umbrel failed to use your storage device in low power mode</b></p>
-        <p>Are you using the recommended hardware?</p>
+        <p><b class="text-dark">Error: Failed to rebind external drive</b></p>
```
lukechilds, line 90: Maybe we can reword from "Failed to rebind external drive" to "Failed to load storage device drivers" or something more simple. I don't think many users will understand what rebinding means or that it relates to drivers.

scripts/umbrel-os/status-server/static/index.html
mayankchhabra: But the status server only runs on Umbrel OS, which runs only Raspberry Pi 4?

mayankchhabra (approved):
The status server works like a dream!!!! Awesome work, @lukechilds 👏

PR #824, opened 2021-06-10
Remove license info 

AnthonyRonning:
It's unnecessary.


AaronDewes:
Removing the license would mean you have no right to use the repo, not even download and use Umbrel except with explicit permission.

HenryWait:
The software license is very necessary, it defines what you can and can not do with the code, such as copying, making changes and redistributing the code. 
Removing or changing the license is usually covered by the license too.

PR #856, opened 2021-06-28
Update Mempool to v2.2.0

bguillaumat:
@wiz I've updated the docker-compose file to be the same as in mempool exemple
I'll let you set the Version/SHA when all is good on your side!


wiz:
@knorrium didn't we rename the docker variables to not mix electrs and electrum

bguillaumat:
> @knorrium didn't we rename the docker variables to not mix electrs and electrum

Yeah right forgot I've made this change 3 months ago, will update now

wiz:
> Yeah right forgot I've made this change 3 months ago

ok well anyway I've just tagged v2.2.0, maybe going forward we should make the PR to update umbrel after we tag new versions, not before :)

bguillaumat:
Oh I must have miswrite my thought, I've made the PR on Mempool to change ELECTRS to ELECTRUM 3 months ago, so, when I've made the PR for umbrel 5 days ago I didnt remember the change I've made on mempool docker :)

junderw:
tACK 4affb10

Works gloriously. Also, inspected the docker images and the hashes matched. (though I guess that would be a problem with docker if they didn't)

wiz:
@lukechilds can you merge this PR ? users are asking for new mempool. thanks

lukechilds (approved):
Thanks guys!

lukechilds:
@wiz we'll get this out in a release this week.

PR #859, opened 2021-06-29
Umbrel v0.3.14

lukechilds:
This update resolves a stability issue that can cause Umbrel to randomly disconnect from the network and become inaccessible.

If you face any difficulties while updating, please message us on Telegram: https://t.me/getumbrel

Diff: https://github.com/getumbrel/umbrel/compare/v0.3.13...v0.3.14


mayankchhabra (approved):
Let's GO 🚀

umbrel-v0 3 14

PR #861, opened 2021-06-30
Add initial PhotoPrism app config

lastzero:
Note:
- HTTP port 2342 will be exposed directly. Production
  environments should use an HTTPS proxy like Traefik in front of
  PhotoPrism. Since it makes sense to set up a single reverse proxy
  for all apps (instead of each app running its own proxy), this
  was left open.
- A random initial admin password should be used instead of a
  static one.
- If PhotoPrism doesn't start it's typically due to wrong storage
  folder permissions, like when it was automatically created by
  Docker as root but the container is running as UID 1000.
- SQLite may be used if MariaDB is too heavy, however it doesn't
  scale well for many pictures and/or cpu cores.

### DIFF
--- a/apps/photoprism/database/.gitignore
+++ b/apps/photoprism/database/.gitignore
@@ -1,0 +1,2 @@
+*
+!.gitignore
--- a/apps/photoprism/docker-compose.yml
+++ b/apps/photoprism/docker-compose.yml
@@ -1,0 +1,86 @@
+version: "3.7"
+
+# Umbrel Docker Compose config file for PhotoPrism (Linux / AMD64)
+#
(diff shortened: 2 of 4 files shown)


mayankchhabra:
Hey @lastzero! With Umbrel 0.5, we moved the all the Umbrel apps to their own dedicated repository — [getumbrel/umbrel-apps](https://github.com/getumbrel/umbrel-apps). With this change, new apps (and app updates) go live in the Umbrel App Store in realtime as soon as the PRs are merged.

I'm closing this PR for now as PhotoPrism is live in our app store (and loved by many). Cheers! :)

PR #864, opened 2021-07-02
Support running bin scripts with argument which contains quotes

### LINKED ISSUE #863: Failed to run `bin/lncli` script
I tried to run the command below, it exited with 1 error.
```bash
$ ./bin/lncli addinvoice --amt=6969 --memo="A coffee for Roger"
$ echo $?
1
```
When I uncomment `set -euo pipefail` in file, it showed the following log:
```
[lncli] unable to parse preimage: encoding/hex: invalid byte: U+006F 'o'
```
And I found that it was caused by quotes in `--memo` argument parsed by`$@` in the script. 

I will do a PR to fix this.

o3o3o:
Fix #863
e.g:
```
./bin/lncli addinvoice --amt=6969 --memo="A coffee for Roger"
```


lukechilds (approved):
Thanks for looking into this @o3o3o!

PR #873, opened 2021-07-10
Remove Tor and Electrs

donaldseaves:
Remove Tor and Electrs


AaronDewes:
@donaldseaves With this fork of Umbrel, you are violating the Umbrel license, so what you are doing is illegal

PR #891, opened 2021-08-06
WIP: Add teslamate

junderw:
Major problems:

1. (**RETRACTED as it's not really that bad. Only bad when compared to https etc.**) Grafana is slow on tor. Extremely slow... 
2. (**SOLVED by removing ports section for home and grafana**) I set tor only as true, but not sure what that does... I can still access via http locally via ip and local hostname. Since this will be a dashboard for personal data and tracking your GPS location for every drive, having that accessible to anyone on your local network seems scary and I'd like to only allow the tor container to have request access to it if possible... in my personal instance I use NGINX Basic Auth password to keep out others... I think "knowledge of the tor onion address" should be enough of a deterrent.
3. (**SOLVED by adding a post install hook**) I need to use a volume for grafana because of an issue in the way they manage sqlite. When I use the normal method, docker creates the folder under 755 root:root permissions. grafana user is 472:0 (it's in the root group(GID) but UID is 472 (name=grafana)) so it chokes on a permission denied error and dies constantly. With volumes, the folder is created (by grafana user) with 777 472:0 permissions... Looking online I see no way around this without grafana container using root user...

TODO:

- Write description
- Take some screenshots (I have a lot of data... so I can take some with my instance.)
- Deal with above problems

When accessing locally, it works great. I can log in with my tesla credentials and it starts pinging my car immediately.

### DIFF
--- a/apps/registry.json
+++ b/apps/registry.json
@@ -1,25 +1,25 @@
 [
     {
-      "id": "nextcloud",
-      "category": "Files",
-      "name": "Nextcloud",
-      "version": "21.0.2",
-      "tagline": "Productivity platform that keeps you in control",
-      "description": "Nextcloud puts your data at your fingertips, under your control. Store your documents, calendar, contacts and photos on your Umbrel instead of some company's data center. Features:\n\n- Mobile and desktop sync \n- Versioning and undelete \n- Galleries and activity feed \n- File editing and preview support for PDF, images, text files, Open Document, Word files and more. \n- Smooth performance and easy user interface. \n- Fine-grained control over access to data and sharing capabilities by user and by group.\n\nNote: After creating your admin account when you open Nextcloud for the first time, you'll be taken to a white (blank) screen. Please close the app and open it again to continue.",
-      "developer": "Nextcloud GmbH",
-      "website": "https://nextcloud.com",
-      "dependencies": [],
-      "repo": "https://github.com/nextcloud/server",
-      "support": "https://help.nextcloud.com/categories",
-      "port": 8081,
-      "gallery": [
-          "1.jpg",
-          "2.jpg",
-          "3.jpg"
-      ],
-      "path": "",
-      "defaultPassword": "",
-      "torOnly": false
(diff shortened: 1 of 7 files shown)


junderw:
Also, is there a way to get the TZ of the user? I set it to UTC as a sane default, but ideally the drive times etc. should be shown in their local time. (Mine is set to "Asia/Tokyo")

junderw:
Also, it's kind of unintuitive since the main web portal and the grafana page don't link to each other by default.
You need to go into the Settings panel of the main web portal and paste in the link to grafana.

I made two discussions / Q&A on the teslamate repo.

https://github.com/adriankumpf/teslamate/discussions/1893
https://github.com/adriankumpf/teslamate/discussions/1892

junderw:
How about a new "post-install.sh" hook?

Run after install.

It could deal with the permissions.

junderw:
I solved 2 and 3 by not exposing the ports and adding a post install script to the app script.

It's slow, but usable, I'm finding. So 1 is meh ok for me.

junderw:
I removed the PORT envs since I wasn't using it anywhere beside the torrc file template.

Since I won't be using the ports externally, having the PORT envs there gives the impression that those PORTs are "taken" to new people adding an app.

junderw:
The decision for how to make this app tor only should also be added to Vaultwarden later on.

Currently you can set up Bitwarden Android app to connect to http://umbrel.local:xxxx and all data is sent in the clear (granted, in your home network, and the password data is locally encrypted, so only encrypted vault data is sent over http.)

mayankchhabra:
Hey @junderw! With Umbrel 0.5, we moved the all the Umbrel apps to their own dedicated repository — [getumbrel/umbrel-apps](https://github.com/getumbrel/umbrel-apps). With this change, new apps (and app updates) go live in the Umbrel App Store in realtime as soon as PRs are merged (you probably know this already haha).

I'm closing this PR for now. If you're still interested in shipping Teslamate, feel free to open a PR on that repo. Cheers! :)

PR #994, opened 2021-09-22
Enable yubikey use with vaultwarden

lvnilesh:


lukechilds:
Of these three PRs:

- https://github.com/getumbrel/umbrel/pull/996
- https://github.com/getumbrel/umbrel/pull/995
- https://github.com/getumbrel/umbrel/pull/994

Are you able to close whichever you are no longer working on?

Also in future, you can just keep pushing to the same branch to update your PR instead of opening multiple PRs.

PR #996, opened 2021-09-22
Enable yubikey use with vaultwarden

lvnilesh:


aphex3k (changes requested):
#994, #995, and #996 should all go in the same PR, no?

PR #1001, opened 2021-09-28
Add IPv6 Support

senepa:
After installing the pi-hole app I discovered that the IPv6 resolution features were unusable. 

As of this post there are thousands of Lightning nodes available on IPv4 space yet less than 200 available via IPv6.
(https://1ml.com/statistics)

This patch aims to provide at least outbound IPv6 support for the pihole application by adding the IPv6NAT container (https://github.com/robbertkl/docker-ipv6nat) and assigning unique local addresses to existing containers.


senepa:
IPv6NAT has been depreciated for native docker IPv6 support (https://github.com/robbertkl/docker-ipv6nat/issues/65)

The following addition in /etc/docker/daemon.json enables IPv6 support
{ "ipv6": true, "fixed-cidr-v6": "fd00::/80", "experimental": true, "ip6tables": true }

PR #1004, opened 2021-09-30
Change squeaknode tagline

yzernik:
This would update the squeaknode tagline. I think the current tagline may confuse users, because they may think that squeaknode uses lightning network itself to send messages between peers.

### DIFF
--- a/apps/registry.json
+++ b/apps/registry.json
@@ -481,7 +481,7 @@
         "category": "Social",
         "name": "Squeaknode",
         "version": "0.1.152",
-        "tagline": "A peer-to-peer status feed on Lightning",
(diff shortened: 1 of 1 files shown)


apps/registry.json
```diff
@@ -481,7 +481,7 @@
         "category": "Social",
         "name": "Squeaknode",
         "version": "0.1.152",
-        "tagline": "A peer-to-peer status feed on Lightning",
+        "tagline": "A peer-to-peer status feed with lightning monetization",
```
lukechilds, line 484:
```suggestion
        "tagline": "A peer-to-peer status feed with Lightning monetization",
```

lukechilds (approved):

lukechilds:
Good idea!

PR #1006, opened 2021-10-02
Allow binding to custom IP address

louneskmt:
Keeping in mind that it is **not recommended** to run Umbrel on a VPS as it is designed to be run on a secure local network, this PR allows VPS users to bind containers ports to a custom IP address (`127.0.0.1` recommended) instead of `0.0.0.0` by default (which opens Umbrel's ports to the world on a VPS connected directly to internet as there is no router).

On first run, `BIND_ADDRESS` can be used to specify the IP address you want to bind Umbrel core services and apps ports to:
```
sudo BIND_ADDRESS=127.0.0.1 ./scripts/start
```

Afterwards, this configuration can be overwritten:
```
sudo BIND_ADDRESS=0.0.0.0 ./scripts/configure
```

Replaces #501


lorenzoPrimi:
Is this gonna be merged? All the VPS umbrel users are waiting!

m-nogues:
Is there a plan to merge this at some point ?

PR #1008, opened 2021-10-02
Clarify BTCPay remote connection behaviour

lukechilds:
Re: https://twitter.com/pavlenex/status/1443977299712528385

@pavlenex sorry about the support issues, does this looks ok?

Added the following note:

>Please note: Your BTCPay instance is only available remotely over the Tor network. Connecting remote applications, such as Shopify or WordPress for example, will fail to connect to your umbrel.local address.
Screenshot 2021-10-02 at 19 03 41


pavlenex (approved):
I think this is good, thanks for a quick fix. 🚀

PR #1010, opened 2021-10-02
Update to electrs v0.9.0

lukechilds:
https://github.com/romanz/electrs/blob/master/RELEASE-NOTES.md#090-sep-30-2021


AaronDewes (changes requested):
`daemon_p2p_addr` should be set in electrs.toml

Also, the update should reset the electrs folder because the db is now different and in a different subdirectory, so if it isn't reset, the old DB is kept around but will stay unused.

You may also need to set `daemon_dir = "/bitcoin"`

PR #1016, opened 2021-10-04
Umbrel v0.4.3

lukechilds:
SECURITY UPDATE: This version of Umbrel brings the latest LND 0.13.3 which includes an important security fix.


mayankchhabra (approved):
🚨 Let's GO!

KayBeSee:
When do you think #1014 will be reverted / new official release pushed? Looking to mess around with electrs 0.9 batching.

PR #1019, opened 2021-10-05
Remove unused configs for squeaknode envoy ip and port

yzernik:
These configs were added during development when squeaknode was still using envoy, but they are unused now.


AaronDewes:
Please also remove it here:

https://github.com/getumbrel/umbrel/blob/4dd77fcf347ab23eb3ca566c5dfa070fe3c29f57/templates/torrc-apps-3-sample#L18

lukechilds (approved):

PR #1092, opened 2021-11-05
Add syncthing app

SamCymbaluk:
# App Submission
## Disclaimer
I am not a Syncthing developer. I am simply packaging it for installation on Umbrel.

### App name
Syncthing

### Version
1.18

**Synchronizes files between two or more computers.**

### Summary of the app
Syncthing is a continuous file synchronization program. It synchronizes files between two or more computers in real time, safely protected from prying eyes. Your data is your data alone and you deserve to choose where it is stored, whether it is shared with some third party, and how it's transmitted over the internet.

Use Syncthing to turn your Umbrel node into a backup server for your important files. You can even configure it to automatically backup your important LND files to a remote phone or computer whenever they change.

### Developer name
The Syncthing Foundation

### Developer website
https://syncthing.net/

### Source code link
https://github.com/syncthing/

### Support link
https://forum.syncthing.net/

### Requires
- [ ] Bitcoin Core
- [ ] Electrum server
- [x] LND

### 256x256 SVG icon
https://reposcope.com/media/unknown/none/scalable/syncthing.svg

### Gallery images
syncthing_homepage
syncthing_settings
syncthing_remotemachine
syncthing_lndfolder

### I have tested my app on:
- [x] [Custom Umbrel install on Linux](https://github.com/getumbrel/umbrel#-installation)


lukechilds:
Thanks for this @SamCymbaluk! I had vaguely heard of Syncthing but wasn't very familiar with it, after palying around with this it is seriously seeeeeriously cool!

apps/syncthing/docker-compose.yml
```diff
@@ -0,0 +1,24 @@
...
+    hostname: umbrel-syncthing
+    environment:
+      - PUID=1000
+      - PGID=1000
+    volumes:
+      - ${APP_DATA_DIR}/data:/var/syncthing
+      - ${LND_DATA_DIR}:/var/lnd:ro
+    ports:
+      - 8384:8384
+      - 22000:22000
```
lukechilds, line 18: Looks like maybe we need to open up port 22000 too?

https://github.com/linuxserver/docker-syncthing/blob/d538b847d3a0035ff0425a9b382d1e555be41440/README.md?plain=1#L89

And maybe also port 21027?

https://github.com/linuxserver/docker-syncthing/blob/d538b847d3a0035ff0425a9b382d1e555be41440/README.md?plain=1#L90

Also looks like the linuxserver image is more popular than the official one. Maybe we should use that instead?

apps/syncthing/docker-compose.yml
```diff
@@ -0,0 +1,24 @@
...
+  # Used to create the marker Syncthing needs to share the lnd data dir.
+  lnd-marker:
+    image: debian:buster-slim
+    command: "mkdir /lnd/.stfolder && exit"
+    volumes:
+      - ${LND_DATA_DIR}:/lnd
```
lukechilds, line 24: Unfortunately we'll have to remove this. An app is not allowed to access another apps filesystem, it's a violation of the app sandbox. We currently allow it for a few apps for LND/Bitcoin Core only so they can read macaroons/cookies but we'll remove that at some point in the future and transition to environment variables.

apps/syncthing/docker-compose.yml
```diff
@@ -0,0 +1,24 @@
+version: "3.7"
+
+services:
+  syncthing:
+    image: syncthing/syncthing
+    restart: on-failure
+    container_name: syncthing
```
lukechilds, line 7: We don't need this.

apps/syncthing/docker-compose.yml
```diff
@@ -0,0 +1,24 @@
+version: "3.7"
+
+services:
+  syncthing:
```
lukechilds, line 4: `server` would be a better name for the service, since it inherits the name of the containg folder as a prefix. So this will currently be `syncthing_syncthing`. `syncthing_server` would be preferable.

lukechilds:
Made some tweaks, good for merge now!

Thanks so much for this @SamCymbaluk! 🚀

SamCymbaluk:
Great! Thanks for making those changes @lukechilds 

Please note that without the lnd-marker service, users will not be able to backup their LND files without manually adding a `.stfolder` folder to the LND directory. I expect many users will want to back those files up, so it might be worth considering alternatives that preserve the app sandboxing without requiring users to SSH into their nodes.

lukechilds:
Yeah the issue is not with the marker service specifically, it's a fundamental issue with one app accessing the file sytem of another app. Apps should not be able to access eachothers file system at all. Currently we are breaking that in some rare cases, simply because the apps currently expect these cookie/macaroon files to exist on disk and don't yet support passing them in via env vars, which can be resolved in future versions.

But in this case the fundamental concept of one app managing files of another app is not something we ever want to allow apps to do. If any app could access the filesystem of any other app there would be significant security implications if a single app was compromised where it could then infect other apps.

If you're familair with how app sandboxing works on iOS/Android, we are working towards a very similar model for Umbrel apps.

shawnyeager:
Could we get an updated, higher resolution logo to go with the rest of the gorgeous Umbrel UI? 🙏🏻

PR #1094, opened 2021-11-06
added option for a domain

SimonHaas:
The only way I found to run umbrel with a domain was to edit the start script and put your custom domain in the DEVICE_HOSTS variable to avoid a CORS error. So I created ​an argument for the script to specify your custom domain without modifying the source.

Caveat: Umbrel dashboard works, except the page /apps. All your apps are being displayed as 'Starting...' and can not be accessed. They are running though and can be accessed via ip:port in your browser.

### DIFF
--- a/scripts/start
+++ b/scripts/start
@@ -1,6 +1,19 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+if [ "$1" == "-h" ]; then
+  cat << EOF
+CLI for starting your Umbrel instance
+
+Usage: [-h] [-u]
+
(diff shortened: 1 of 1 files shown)


eddy-geek:
Just a amall note, https://github.com/getumbrel/umbrel/pull/708 is also related to CORS and is touching same lines.

mayankchhabra:
Sorry for the delay, Simon. And thanks for working on this. The CORS check was updated a while ago and this should be a non-issue. My apologies for the resulting unproductive work.

PR #1102, opened 2021-11-10
Add SatSale App

nickfarrow:
### App name
SatSale

### Version
2.0.1

A lightweight Bitcoin payment processor.

### Summary of the app
SatSale is a simple, easily deployable, lightweight Bitcoin payment processor that connects to your own Bitcoin node. You can use SatSale to instantly turn your Umbrel into a point-of-sale (settings under `/admin`). SatSale can process donations from your supporters, facilitating direct peer-to-peer payments without any middleman and greater privacy than reusing donation addresses.

Host your own Bitcoin payment gateway for Woocommerce without a middleman or custodian, allowing you to turn any wordpress site into a bitcoin-accepting store.

This Umbrel app contains a subset of the features we have in the works for SatSale, but will push updates out as they are ready. Check out the website and GitHub for more!

Please note that SatSale is still in early development. As such, we are not responsible for any loss of funds, vulnerabilities, or any other grievances which arise. Always confirm large payments manually and use cold storage as much as possible. Please contribute if you can!

### Developer name
Nick Farrow & SatSale Developers

### Developer website
https://satsale.org

### Source code link
https://github.com/nickfarrow/SatSale/

### Support link
https://github.com/nickfarrow/SatSale/issues

### Requires
- [X] Bitcoin Core
- [ ] Electrum server
- [X] LND

### 256x256 SVG icon
https://svgshare.com/i/bhk.svg

### Gallery images
**Updated**
https://imgur.com/a/OKoQyEn

### I have tested my app on:
- [X] [Umbrel dev environment](https://github.com/getumbrel/umbrel-dev)
- [X] [Umbrel OS on a Raspberry Pi 4](https://github.com/getumbrel/umbrel-os)


nickfarrow:
A few things:

- Let me know if you need anything else, I can get some better screenshots.
- I've noticed a bug where the js keeps making requests after a payment is confirmed, fixing this. Also let me test it is working with woocommerce payment gateway,

Some questions I have

1. Is there a way to point a domain to an Umbrel?
2. At the moment i'm using a pretty inaccessible config.py which is probably not ideal for Umbrel (particularly for some other features I could push like lnaddress and auto swap for LUSDT options - they need a config line. Is there an easy way without having to build a config page (can do but would take some time, need password etc)?
3. Similarly, the SatSale API key for woocommerce gateways is printed to trouble shooting logs, not very accessible. Perhaps a config page is the way to go..

Thanks again!

nickfarrow:
I've fixed a bug that was causing repeated check_payment requests after a payment was complete, also I checked and everything is working with the Woocommerce plugin (tested by pointing to IP).

yzernik:
Cool app!

I was having similar issues with Python config in my app, so I created a class that reads config from:
1) Environment variables (high priority)
2) INI file (low priority)

https://github.com/yzernik/squeaknode/blob/7e848c3c39a93dd62249c454a1b420e43507bd21/squeaknode/config/config.py#L170-L194

You can copy any of it if you like.

lukechilds:
Hey @nickfarrow, so sorry for the delay! Thing have been crazy. Reviewing this now.

As for your questions:

>Is there a way to point a domain to an Umbrel?

Unfortunately there's not a simple way to do that currently. We also don't reccomend users manaully go setting it up themselves since Umbrel is currently in beta and is not yet ready to be exposed publically to the internet. Currently Umbrel should either be used over the local network or remotely via Tor, making sure to only share the Tor hidden service with people you trust.

>At the moment i'm using a pretty inaccessible config.py which is probably not ideal for Umbrel (particularly for some other features I could push like lnaddress and auto swap for LUSDT options - they need a config line. Is there an easy way without having to build a config page (can do but would take some time, need password etc)?

Yeah if you wan't users to be able to change these values you'd need to provide some way for them to do this via the web ui and of course authenticate that. We do pass an environment variable to all apps `$APP_PASSWORD` which is a 128 bit cryptographically secure random hex string unique to each app which you can use as a default password value. It will be shown to the user in the Umbrel UI for this app so they can easily copy/paste it over.

>Similarly, the SatSale API key for woocommerce gateways is printed to trouble shooting logs, not very accessible. Perhaps a config page is the way to go..

Yeah you'd need to show this to users in the web ui for them to be able to copy it over.

apps/satsale/docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+    image: satsale/satsale:0.2.9
+    restart: on-failure
+    stop_grace_period: 1m
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+      - ${LND_DATA_DIR}:/lnd:ro
+    ports:
+      - 5000:5000
+    environment:
+      DATA_DIR: ${APP_DATA_DIR}
```
lukechilds, line 14: This should be the data dir inside the container not on the host.
```suggestion
      DATA_DIR: "/data"
```

apps/satsale/docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+      # Bitcoin Core connection details
+      BITCOIND_HOST: $BITCOIN_IP
+      BITCOIND_PORT: $BITCOIN_RPC_PORT
+      BITCOIND_USER: $BITCOIN_RPC_USER
+      BITCOIND_PASS: $BITCOIN_RPC_PASS
```
lukechilds, line 19: Are these paramters required or is everything done via LND now? If they are done via LND we should remove them, if they are still required what are they used for?

apps/satsale/docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+      LND_DATA_DIR: $LND_DATA_DIR
+      LND_HOST: $LND_IP
+      LND_GRPC_PORT: $LND_GRPC_PORT
+      MACAROON_FILE: /lnd/data/chain/bitcoin/${BITCOIN_NETWORK}/admin.macaroon
```
lukechilds, line 23: Is the admin macaroon required by SatSale or can it use the invoice macaroon?

lukechilds (changes requested):
This looks really cool, a few changes needed but pretty much good to merge.

Also I noticed you link to https://github.com/nickfarrow/SatSale in the web ui, that's fine, but you should set the referrer policy to `no-referrer` to avoide Tor users leaking their hidden service to you/GitHub. You can do this by adding a simple HTTP header or HTML meta tag: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy

apps/satsale/docker-compose.yml
nickfarrow: Good point, removing

apps/satsale/docker-compose.yml
nickfarrow: Let me try get invoice.macaroon working, think I was having some issue before.

nickfarrow:
Ok I'll I'll see how I go with a settings page behind some authentication. Also I am thinking of pulling the features from the SatSale [point-of-sale branch](https://github.com/nickfarrow/SatSale/tree/webstore), but might leave for a later update.

lukechilds:
Hey @nickfarrow what's the status on this from your end?

Do you want to implement the settings/auth before shipping? Or are you happy shipping SatSale to the Umbrel app store as it is currently and providing those features in an update?

nickfarrow:
Hey @lukechilds , just got back from Bitcoin bushbash and haven't had much time to make the settings page. Working on it now and it is my main priority this week. Approximately when is the next update?

My concern is that if we ship without a settings page, then a donation page that isn't exposed to the internet isn't very useful. An instant point-of-sale would be however.

lukechilds:
Got it.

Aiming for an update very soon this week but if you can't make it we'll get you in the next one.

Keep us posted!

lukechilds:
Hey @nickfarrow how are you getting on?

We're aiming for another release before the end of the year, do you reckon you'll have SatSale in a state you're happy shipping it in before then?

nickfarrow:
Hey @lukechilds , really sorry for the slow progress with this. Juggling a few things at the moment. 

I've got a docker image that has an `/admin` behind basic password authentication that allows for changing the `/store` items. Testing it on my umbrel now, then going to make it look a little nicer and it should be ready to publish.

lukechilds:
@nickfarrow sounds great!

>BTW is there any easy way to find the $APP_PASSWORD on my umbrel prior to it being a real app?

It just exists as an envvar that gets passed to your app. You can add:

```yaml
environment:
  APP_PASSWORD: $APP_PASSWORD
```

to your app and you should be able to use it.

You should be able to run `./scripts/app compose satsale exec satsale env` to execute `env` inside your app container and dump all it's env vars which will include `$APP_PASSWORD`.

>sudo scripts/update/update downgrades me to 4.08, is this normal?

I'm not sure I understand. What's the full command you are running and what are you expecting it to do?

`scripts/update/update` does a few different things but the common use case is to pass it a git repo and branch to update to. e.g `sudo ~/umbrel/scripts/update/update --repo nickfarrow/umbrel#satsale` will migrate an Umbrel to this PR branch. It should only be run on a full Umbrel install and not umbrel-dev. It also should only be run for testing and not on personal nodes since it skips a few safety checks that the OTA updates do when started from the UI.

nickfarrow:
I've added the no-referrer to links and cleaned up the point-of-sale settings a little bit. The point-of-sale is functional, but could still do with some work, particularly for touchscreen tablet & mobile (I think this app will attract some new contributors 😁).

The settings is behind basic auth loging into the `/admin` page using the $APP_PASSWORD, and also for submitting forms (can add cookie system or something later).

I've also taken some new screenshots. Sorry i've bumped the version like 20 times..

lukechilds:
Awesome! Is this ready for merge now?

nickfarrow:
Yep good to go! Could you please have one last play around for me?

Thanks again, very excited to get this out!

lukechilds (changes requested):
Are you sure you added the no referrer header? I don't see it set in the response, and I can see the referrer header being set in all external requests links.

Also there appears to be a bug in the authentication. It looks like it gets validated after processing inbound requests instead of before. e.g I can do `curl -F file=@/some/huge/file http://umbrel.local:5000/uploader` and `/some/huge/file` get uploaded to the server despite us not setting a password. If I do that with a 1GB file it hangs a core at 100%, prevents SatSale from accepting any requests, before killing the worker thread and restarting it. If you shared your SatSale link with anyone or embedded it in a website, anyone would be able to perform this attack.

This is a DoS attack vector to both SatSale and Umbrel in general. If you validate auth before handling the request you can prevent this. You could also add a max limit to the filesize as a sanity check.

nickfarrow:
I have fixed the noreferrer on all links. 

I wasn't able to reproduce the upload without password issue, I just get an internal 500 error because it is missing the `pass` field. I did however add a file size limit (1MB, could be lowered). 
https://github.com/nickfarrow/SatSale/commit/971fc0d551c20925435c6ea8d9a8fa5716019e2a#diff-5df2490095b1806f64debd7bbbfa453108d824ebd0c390ae5fe760f52d614344R35

In this snippet, I am checking for the password prior to `save(file)`, but maybe the upload already happens along with the request, with this library. I will look more tomorrow. Thanks

nickfarrow:
In that latest version, my curls are getting rejected for invalid password or internal server error if not present. It doesn't appear that the file upload begins.

lukechilds:
Awesome! Yep just confirmed with the latest version I get an instant 413 with a large payload instead of the sustained CPU usage and crash I was seeing before.

Quick Q, do you think the Umbrel UI should link to the SatSale root url or `/admin`? I'm thinking `/admin` makes more sense since the admin link may not be obvious to Umbrel users and it isn't linked to anywhere from the root url. What would you prefer?

lukechilds:
Merged, thanks so much @nickfarrow!

nickfarrow:
fantastic, thanks so much Luke for all your help

louneskmt:
Hey @nickfarrow! Some users reported the generation of two 100 000 000 sats invoices right after the installation of SatSale. Do you have an idea where this come from?

nickfarrow:
Hey @louneskmt ,
I think it's coming from here where we test connection to lnd by creating an invoice.
https://github.com/nickfarrow/SatSale/blob/docker/node/lnd.py#L43

I could delete this invoice, but should probably replace it with some other call `invoice.macaroon` has access to, to ensure lnd is ready.

https://github.com/nickfarrow/SatSale/issues/35

louneskmt:
> I think it's coming from here where we test connection to lnd by creating an invoice.

Thanks, good to know.

> I could delete this invoice, but should probably replace it with some other call

Yeah, that would be better, because the pending invoices are displayed in the UI and it confuses users (and they ask themselves if their Umbrel has been compromised).

lukechilds:
Hey @nickfarrow any update on this?

We've had a number of reports from Umbrel users who are worried their Umbrel has been compromised after seeing an invoice show up they didn't create. Maybe it would be enough to just do the connection check via `GetInfo` or something less invasive?

nickfarrow:
Yep ill try make another release today

nickfarrow:
Fixed in [0.1.13](https://hub.docker.com/layers/satsale/satsale/0.1.13/images/sha256-22fb2018e8125a546645512ce911bd203a8ef50a40bd0c8b83e36cce79b7273a?context=repohttps://hub.docker.com/layers/satsale/satsale/0.1.13/images/sha256-22fb2018e8125a546645512ce911bd203a8ef50a40bd0c8b83e36cce79b7273a?context=repo)

PR #1091, opened 2021-11-04
Update LNMarkets to v1.1.3

kibotrel:
**Patchnotes:**
 - [v1.1.0](https://github.com/ln-markets/umbrel/releases/tag/v1.1.0)
 - [v1.1.1](https://github.com/ln-markets/umbrel/releases/tag/v1.1.1)
 - [v1.1.2](https://github.com/ln-markets/umbrel/releases/tag/v1.1.2)
 - [v1.1.3](https://github.com/ln-markets/umbrel/releases/tag/v1.1.3)

PR #1105, opened 2021-11-10
Revert "Update LNMarkets to v1.1.3"

lukechilds:
Reverts getumbrel/umbrel#1091

Failing with

```
Error response from daemon: Get "https://ghcr.io/v2/lnmarkets/umbrel/manifests/sha256:7f72864171bbe7922929d940a63c31faed6aa7525b6a236f5a4e0d2680ccdeba": denied
```

### DIFF
--- a/apps/lnmarkets/docker-compose.yml
+++ b/apps/lnmarkets/docker-compose.yml
@@ -2,7 +2,7 @@ version: '3.7'
 
 services:
   lnmarkets:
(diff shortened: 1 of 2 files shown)


lukechilds:
Doesn't seem to be an issue with the update, the old container is giving the same error.

lukechilds:
cc  @kibotrel looks like your images are no longer available:

```
$ docker pull ghcr.io/lnmarkets/umbrel
Using default tag: latest
Error response from daemon: Head https://ghcr.io/v2/lnmarkets/umbrel/manifests/latest: denied
```

kibotrel:
Seems like a little issue with our organization name we went from `lnmarkets` to `ln-markets` between the two releases and forgot to update it!

PR #1111, opened 2021-11-15
Umbrel v0.4.8

lukechilds:
Umbrel 0.4.8 brings a new security feature — unique cryptographically secure default app passwords derived from your 24 secret words. This will help protect your Umbrel even if an app's unique Tor URL gets leaked and you have not changed that app's default password.

If you have currently installed ThunderHub, Lightning Terminal, Ride The Lightning, Squeaknode or Code Server on your Umbrel, their default passwords will automatically upgrade to the newer, more secure passwords which can be found on their app store listing pages. In case you have already updated passwords of these apps manually, you can continue to use the same.


mayankchhabra (approved):
awesome work my man, pure fire today 🔥

PR #1115, opened 2021-11-18
Update ride-the-lightning app entrypoint to use sh

### LINKED ISSUE #1112: RTL won't start after updating to 0.4.8
first installed update over web ui, RTL says starting but nothing can be reached on port 3001
tried to:
- uninstall RTL reinstall
- stop, manually update umbrel with curl, start
- stop, docker container prune, docker image prune, start

same symptoms

louneskmt:
RTL Docker image for amd64 uses node alpine as base, which doesn't include `bash`. Changing `bash` to `sh` should solve the issue of RTL not starting for Umbrel setups on amd64. 

Resolves #1112.

### DIFF
--- a/apps/ride-the-lightning/rtl/entrypoint.sh
+++ b/apps/ride-the-lightning/rtl/entrypoint.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 
 # Migrate legacy default password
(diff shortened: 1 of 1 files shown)


aphex3k (approved):

louneskmt:
Tested on both amd64- and arm64-based devices (by @jairunet and myself), so it is good to be merged @lukechilds.

lukechilds:
Thanks for looking into this!

PR #1117, opened 2021-11-19
Update squeaknode website

yzernik:
Update the registry to point to the squeaknode website.


mayankchhabra (approved):
Nice website, @yzernik! 🎉

PR #1123, opened 2021-11-24
Add boltz service to Ride The Lightning app

aphex3k:
## Provide [boltz](https://github.com/BoltzExchange/boltz-lnd) service/container to RTL app.

This PR seeks to configure boltz service in ride-the-lightning app

### Pro

- Currently, RTL only connects to its loop container. The Boltz service is not configured. 
- Loop In/Out is only available for channels with a minimum capacity of 250K satoshis. Boltz submarine swap in/out allows channels as small as 50K satoshis capacity.

### Con
Unfortunately the Boltz project only provides the amd64 flavor of boltz-lnd which would not run on arm-based umbrel installations like any Raspbery Pi or umbrel OS running on M1 macs. Thus I used `buildx` to create both images and pushed them to my own dockerhub. Once boltz-lnd provides both image flavors on their own dockerhub we can switch over to theirs.


aphex3k (commented):
Alternatively the boltz-lnd container could be moved to the RTL apps docker composition file.

aphex3k (commented):
> Alternatively the boltz-lnd container could be moved to the RTL apps docker composition file.

I did just that. This PR should fit a lot better now.

lukechilds:
Thanks for working on this @aphex3k! Unfortunately we can't have the image built as part of the installation step, we'd need Boltz to support multiarch Docker containers before accepting this PR.

aphex3k (commented):
> Thanks for working on this @aphex3k! Unfortunately we can't have the image built as part of the installation step, we'd need Boltz to support multiarch Docker containers before accepting this PR.

@lukechilds Done. I used `buildx` to create both images and pushed them to my dockerhub. Once boltz-lnd provides both image flavors on their own dockerhub we can switch over to theirs.

lukechilds:
This looks great @aphex3k!

Would be really great to use the official image, I see they use GitHub Actions [here](https://github.com/BoltzExchange/boltz-lnd/blob/master/.github/workflows/artifacts.yml), maybe we could open a PR adding your Docker command here and pushing to their Docker Hub?

aphex3k:
> This looks great @aphex3k!
> 
> Would be really great to use the official image, I see they use GitHub Actions [here](https://github.com/BoltzExchange/boltz-lnd/blob/master/.github/workflows/artifacts.yml), maybe we could open a PR adding your Docker command here and pushing to their Docker Hub?

I'm not familiar with Github Actions but I will educate myself...

aphex3k:
@lukechilds This PR is now using the boltz images from Dockerhub

lukechilds:
Thanks a lot for this @aphex3k!

PR #1128, opened 2021-11-26
Enable accepting Atomic Multipath Payment (AMP)

aphex3k:
Since v13.0 lnd supports AMP between paying and receiving lnd nodes. 

For umbrel nodes to support receiving amp, they need to be enabled in `lnd.conf`. This patch allows for both `amp` and `keysend` transactions to be received by the node. 

See: https://docs.lightning.engineering/lightning-network-tools/lnd/amp

### DIFF
--- a/templates/lnd-sample.conf
+++ b/templates/lnd-sample.conf
@@ -13,6 +13,7 @@ tlsextradomain=<hostname>
 tlsautorefresh=1
 tlsdisableautofill=1
 feeurl=<lndfeeurl>
+accept-amp=1
 
 [Bitcoind]
 bitcoind.rpchost=<bitcoin-ip>


jasonpearce:
Since LND is already at version 13, could a user simply enable Atomic Multi-path Payments (AMP) by updating their lnd.conf file?

- SSH into your node
- cd ~/umbrel/lnd
- sudo nano lnd.conf
- Add "accept-amp=1" in your lnd.conf file
- Save and Exit
- Reboot or Restart LND

Should that work?

rafftyl:
Is there anything else needed for this to be merged? It's a tiny change, but it would solve a lot of problems related to big payments. I know that we can modify the config on our own, but then we'll have to remember to reapply the change after each Umbrel update.

lukechilds:
>I know that we can modify the config on our own, but then we'll have to remember to reapply the change after each Umbrel update.

@rafftyl changes you make to lnd.conf are persisted. If you add `accept-amp=1` it will stay there after app updates and Umbrel updates.

>Is there anything else needed for this to be merged?

We'll need to make the change over in the [getumbrel/umbrel-apps](https://github.com/getumbrel/umbrel-apps) repo since the Lightning app has migrated there. We have some updates planned for the Lightning app shortly so we can add this as part of the next update.

rafftyl:
@lukechilds 
> changes you make to lnd.conf are persisted.
Oh, that's cool. Thanks for the info.

nevets963:
Closing as this was addressed in this commit: https://github.com/getumbrel/umbrel-apps/commit/7d7b37f02e2b7ebdc70b7ad4460151f410fe4fd6

lukechilds:
amp is now enabled by default in the latest version of the Lightning app, thanks for your suggestion @aphex3k!

PR #1143, opened 2021-12-04
update home-assistant to 2021.11.5 and expose USB device

jroulin:


apps/home-assistant/docker-compose.yml
```diff
@@ -2,7 +2,9 @@ version: "3.7"
 
 services:
   server:
-    image: homeassistant/home-assistant:2021.9.6@sha256:3b0b1d1e595a77fdca2e765d4441545ab873841fc311964f8d6cf9cb9e99f38e
+    image: "ghcr.io/home-assistant/home-assistant:2021.11.5@sha256:1091e8745acbefe1dfa170748d5bf02e5591aff80cc763c8694b030dfb51cf15"
+    devices:
```
lukechilds, line 6: Unfortunately we can't pass devices in like this since it breaks the app sandbox. We have some ideas on how we can give certain apps hardware access without major sandbox implications but they unfortunately aren't ready yet.

apps/home-assistant/docker-compose.yml
```diff
@@ -2,7 +2,9 @@ version: "3.7"
 
 services:
   server:
-    image: homeassistant/home-assistant:2021.9.6@sha256:3b0b1d1e595a77fdca2e765d4441545ab873841fc311964f8d6cf9cb9e99f38e
+    image: "ghcr.io/home-assistant/home-assistant:2021.11.5@sha256:1091e8745acbefe1dfa170748d5bf02e5591aff80cc763c8694b030dfb51cf15"
```
lukechilds, line 5: Are you able to change this back to the Docker Hub registry URL?

lukechilds:
Thanks for this @jroulin! Left some comments.

apps/home-assistant/docker-compose.yml
jroulin: Is there a way to customize this by myself like a hook that is called on each upgrade if we can't do it for everybody? Else home-assistant will not work for me after every upgrade.

apps/home-assistant/docker-compose.yml
jroulin: Yes I can but it was the image recommended by the project. Why is it better to go to docker hub?

jroulin:
closing as it is now obsoleted by https://github.com/getumbrel/umbrel/pull/1160

PR #1146, opened 2021-12-06
App Submission: Lightning Shell

ibz:
# App Submission

### App name

Lightning Shell

### Version

v0.1.1

Web shell with a selection of LN node management utilities

### Summary of the app

Lightning Shell is a web shell for the Umbrel personal server.

Technically it's a Docker container with ttyd and some additional utilities.
* `lncli`
* [`charge-lnd`](https://github.com/accumulator/charge-lnd)
* [`lntop`](https://github.com/edouardparis/lntop)
* [`rebalance-lnd`](https://github.com/C-Otto/rebalance-lnd)
* [`suez`](https://github.com/prusnak/suez)

### Developer name

Ioan Bizău

### Developer website

https://ibz.me

### Source code link

https://ibz.github.io/lightning-shell/

### Support link

https://ibz.github.io/lightning-shell/issues

### Requires
- [x] Bitcoin Core
- [x] LND

### 256x256 SVG icon
https://ibz.me/wesh.svg

### Gallery images

s1
s2
s3

### I have tested my app on:


lukechilds:
Apologies for the delay on this @ibz aiming to review this this week!

apps/wesh/docker-compose.yml
```diff
@@ -0,0 +1,16 @@
...
+    image: ibz0/wesh:v0.2.1
+    restart: on-failure
+    stop_grace_period: 1m
+    ports:
+      - 7681:7681
+    volumes:
+      - ${LND_DATA_DIR}:/lnd:ro
+    environment:
+        LND_HOST: "${LND_IP}"
+        LND_ADDRESS: "//${LND_IP}:10009"
```
lukechilds, line 14: Are both LNDHOST and LND_ADDRESS needed? Seems like LND_HOST is redudant.

apps/wesh/docker-compose.yml
```diff
@@ -0,0 +1,16 @@
...
+        CERT_PATH: "/lnd/tls.cert"
+        MACAROON_PATH: "/lnd/data/chain/bitcoin/mainnet/readonly.macaroon"
```
lukechilds, line 16: Is the readonly macaroon sufficient to work with all the included tools? Don't they need to be able to spend to rebalance?

apps/wesh/docker-compose.yml
```diff
@@ -0,0 +1,16 @@
+version: "3.7"
+
+services:
+  web:
+    image: ibz0/wesh:v0.2.1
+    restart: on-failure
+    stop_grace_period: 1m
+    ports:
+      - 7681:7681
+    volumes:
```
lukechilds, line 10: I notice you haven't added any volumes to persist data for the installed tools. Are they all ephemeral are do some of them expect data to be persisted? If they need data persisted they should be defined here.

apps/wesh/docker-compose.yml
lukechilds, line 14: We should use `$LND_GRPC_PORT` here:

```suggestion
        LND_ADDRESS: "//${LND_IP}:${LND_GRPC_PORT}"
```

lukechilds (changes requested):
Just had a play with this @ibz, super cool! Something like this is definitely needed on Umbrel!

One issue, there's currently no auth. That means anyone on the local network or who gets access to the apps Tor address can see a lot of information about the node. We should add some auth to the app. Luckily both Umbrel and ttyd have built in functionality to handle this so it should be simple.

Umbrel passes an `$APP_PASSWORD` env var to the app to use which will be shown to the user in the dashboard. And ttyd accepts a `--credential USER:PASSWORD` flag to add basic HTTP auth.

Also, it might be a bit confusing for new users to know which tools are installed unless they keep referring back to the app store description. Maybe we could add a bash MOTD that prints a brief message on which tools are installed and what they do. What do you think?

Lastly, what about changing the name from "Wesh" to "Lightning Shell"? Now that we have non-bitcoin apps in the app store the name could be a little unclear. Totally up to you of course.

apps/wesh/docker-compose.yml
lukechilds: Ahh these are for seperate apps, right?

Maybe this would be better handled by a wrapper in the Docker container. So wesh just accepts the minimum args it needs, and then the app specific args are built inside the container.

ibz:
Thanks for the review! Good points, I'll address them.

About the name - I was thinking that this does *not* have to be exclusively LN related, especially since Umbrel is trying to be more than just a BTC node. I just happened to add some LN utilities, but I was meaning to add more. There's a lot of amazing command line utilities that are useful on Umbrel. It's just a lot of work and I started with something I was interested in at the moment, but there should be more in the future.

ibz:
I think I covered all your points. Happy new year, BTW!

ibz:
@lukechilds Oh, BTW, I figured the reason why you proposed a name change was so you could include it in the "LN Management" group in the app store? Otherwise it doesn't really fit in any other group? I guess in that case I could rename it to something like "Lightning Shell" for now. Later it would probably fit into an "Utilities" group though...

apps/wesh/docker-compose.yml
```diff
@@ -2,15 +2,16 @@ version: "3.7"
...
       - 7681:7681
     volumes:
       - ${LND_DATA_DIR}:/lnd:ro
+      - ${APP_DATA_DIR}/data:/data
     environment:
-        LND_HOST: "${LND_IP}"
-        LND_ADDRESS: "//${LND_IP}:10009"
-        CERT_PATH: "/lnd/tls.cert"
-        MACAROON_PATH: "/lnd/data/chain/bitcoin/mainnet/readonly.macaroon"
+        WESH_VERSION: "v0.3.2"
```
lukechilds, line 14: Maybe this would be better off being set in the Dockerfile so it is linked to the actual version rather than being passed in dynamically at runtime?

lukechilds:
Happy new year!

>I guess in that case I could rename it to something like "Lightning Shell" for now.

Sounds great!

lukechilds (approved):
Awesome stuff @ibz, this is badly needed, I think you're gonna make a lot of node operators very happy in the next update!

PR #1162, opened 2021-12-12
Add automatic decrypt and restore SCB script

louneskmt:
This PR integrates into Umbrel codebase a script I've been using with users for a couple of weeks. It reads the backup from a specified path and automatically decrypts the file if necessary (i.e. if it is an automatic Umbrel backup, PGP encrypted), and restores the SCB.

This is temporary, as we'll be moving LND into the App Store at some point.

### DIFF
--- a/scripts/backup/restore
+++ b/scripts/backup/restore
@@ -1,0 +1,88 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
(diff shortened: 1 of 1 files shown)


lukechilds:
Great stuff @louneskmt!

PR #1164, opened 2021-12-12
Update squeaknode repo URL and support URL

yzernik:
The squeaknode repo migrated to the squeaknode github organization, so there is a new URL.


lukechilds:
Thanks for the update Jonathan!

PR #1165, opened 2021-12-13
Umbrel v0.4.10

lukechilds:
Umbrel 0.4.10 brings a brand new app to the Umbrel App Store — Uptime Kuma, support for Boltz swaps in Ride The Lightning app, updated apps including the latest Lightning Terminal app which contains an important security fix, bug fixes, and more.


mayankchhabra (approved):
let's goo 🚀

PR #1178, opened 2022-01-02
App update: helipad to v0.1.6

daveajones:
Two bug fixes:  one caused a gap in the boostagram list and the other could cause the boostagram list to be blank if an unsettled invoice became the current invoice index.


drebscott:
Looking forward to this update and the app being published! :)

lukechilds:
Thanks @daveajones!

PR #1182, opened 2022-01-04
Update APP_SEED description

### LINKED ISSUE #1180: `APP_SEED` is a 32 bytes hex string (256 bit) and not 16 bytes (128 bit)
The documentation says that `APP_SEED` is a 128 bit hex string: 

https://github.com/getumbrel/umbrel/blob/9f34ed7c62666228ce3e4842ebeed0dadf0453d6/apps/README.md?plain=1#L155

However, in the code a sha256 is used to generate `APP_SEED` which ends up in 32 bytes (256 bit): 

https://github.com/getumbrel/umbrel/blob/2f6e5357eb9248d647004aa10e35d46705f4e90d/scripts/app#L56

Am I missing something or is the documentation wrong?

bonomat:
`APP_SEED` is derived deterministically from the master Umbrel seed which contains 128 bits of entropy. While the hex string APP_SEED is 256 bits long, it only provides 128 bits of entropy.

resolves #1180

### DIFF
--- a/apps/README.md
+++ b/apps/README.md
@@ -152,7 +152,7 @@ services:
       # $APP_HIDDEN_SERVICE - The address of the Tor hidden service your app will be exposed at
       # $APP_DOMAIN - Local domain name of the app ("umbrel.local" on Umbrel OS)
       # $APP_PASSWORD - Unique plain text password that can be used for authentication in your app, shown to the user in the Umbrel UI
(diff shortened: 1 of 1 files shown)


lukechilds:
I added a clarification for the entropy @bonomat, do you think it's more clear or too verbose? I just don't want to say 256 bits and let users assume that means 256 bits of entropy when in actual fact they are dealing with much less entropy than they thought they were.

bonomat:
Thanks. Imho this is understandable.

PR #1185, opened 2022-01-06
Upgrade ItchySats app to `0.3.6`

scratchscratchscratchy:
ItchySats now uses Umbrel's `$APP_SEED` to initialize the app's Bitcoin wallet.

Would be great if we could get this in before the Umbrel release, because it is a breaking change to the ItchySats Bitcoin wallet.
Given this is the first Umbrel release with ItchySats we added the feature to use `$APP_SEED` for deriving the wallet now, rather than upgrading a wallet that was created with a generated seed.

@lukechilds you tested opening a position. In case you still have funds in the app's wallet, when upgrading to this version please first withdraw all funds from the Bitcoin wallet and then transfer them back into the wallet derived from Umbrel's `$APP_SEED` once upgraded.

### DIFF
--- a/apps/itchysats/docker-compose.yml
+++ b/apps/itchysats/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3.7"
 
 services:
   web:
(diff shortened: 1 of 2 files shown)


lukechilds:
Tested and working great! Tested transfering some funds, then nuking the app and re-installing. The funds were sucessfully recovered 🎉

PR #1193, opened 2022-01-16
Update the mempool app to v2.3.0

knorrium:
Read the full changelog [here](https://github.com/mempool/mempool/releases/tag/v2.3.0)

This release includes changes to the Docker environment variables which are documented on the mempool [top-level README](https://github.com/mempool/mempool/blob/master/README.md#docker-installation). It also handles the initial tables creation and database migration through the backend, so the SQL file is no longer needed.


wiz:
ACK

lukechilds:
Thanks for this guys, congrats on the 2.3.0 release! Tested locally and working great!

We currently have `deprecatedrpc=addresses` in [bitcoin.conf](https://github.com/getumbrel/umbrel/blob/af61e2df139d5d7a01cd0afecfa7d6157b0e7e0d/templates/bitcoin-sample.conf#L37), am I correct in assuming this is no longer needed in this version of mempool?

Also, just reading the docs, [this](https://github.com/mempool/mempool/blob/master/README.md#bitcoindelectrum-configuration) suggests that `MEMPOOL_BACKEND=electrum` should be set, however it's not set in this PR but the connection seems to work ok. Do we need to add it?

wiz:
> We currently have deprecatedrpc=addresses in bitcoin.conf, am I correct in assuming this is no longer needed in this version of mempool?

Yes IIRC that workaround was only needed for mempool v2.2 + Bitcoin v22

> Also, just reading the docs, this suggests that MEMPOOL_BACKEND=electrum should be set, however it's not set in this PR but the connection still seems to work ok. Do we need to add it?

Since it's the default value I guess it's not needed to set this in the docker-compose but wouldn't hurt to set it anyway

wiz:
FYI the default value of `MEMPOOL_BACKEND=electrum` gets set here:
https://github.com/mempool/mempool/blob/v2.3.0/docker/backend/start.sh#L5

lukechilds (approved):
LFG! 🚀

PR #1204, opened 2022-01-24
Update Krystal Bull

user411:
Adds authentication to Krystal Bull and a header bar for the app, an about route, and rolls up bitcoin-s-oracle-server version.

There was some refactoring under the hood to use our common-ts library and move NodeJS handlers out into middleware files that are common with the Suredbits Wallet, but functionally the main application changed very little.


lukechilds:
This looks great! Tested and working perfectly.

A couple of minor tweaks required, doesn't look like I have permission to push to this branch, can you merge this PR?

https://github.com/bitcoin-s/umbrel/pull/5

lukechilds:
Just to explain the fixes, if you pass in the `log` dir without first creating it, Docker will create it as root because it doesn't exist. Becuase your container runs as uid 1000 it won't be able to write to it. We need to add it to the data dir skeleton (https://github.com/bitcoin-s/umbrel/pull/5/commits/e61f7946a8a376129a02c131f88ccf35c28d64f2) so new users have it created with the correct permissions. We also add a migration step (https://github.com/bitcoin-s/umbrel/pull/5/commits/2cb6bd8826a0ce5d0f946b671d6dab5f74c19c72) so users who already have the app installed get the new dir created on update.

We need to handle this better in Umbrel and expose a proper way for apps to update the filesystem, but for now we're just hardcoding the fixes.

And since you're now using the password feature, we need to set `deterministicPassword` to true in the registry file (https://github.com/bitcoin-s/umbrel/pull/5/commits/def48d34f5e11bcff3f4788ae8a270e828896d62) so Umbrel displays the password to users in the UI.

lukechilds:
Oh also just for future reference, if you submit PRs form a feature branch instead of master, you should be able to check "Allow edits from maintainers" so I can push changes directly to your PR branch. It makes it a bit simpler to do PR review, but no worries if you prefer not to give access, I can just submit PRs like this.

lukechilds:
Once these changes are merged in this PR is good to be merged 🚀

lukechilds:
Merged, thanks guys!

PR #1205, opened 2022-01-26
update sphinx-relay to 2.2.5

Evanfeenstra:
- liquid-based badges on people page
- start LSAT integration
- many more tests

### DIFF
--- a/apps/sphinx-relay/docker-compose.yml
+++ b/apps/sphinx-relay/docker-compose.yml
@@ -3,7 +3,7 @@ version: "3.7"
 services:
   sphinx-relay:
     container_name: sphinx-relay
-    image: sphinxlightning/sphinx-relay:v2.2.3@sha256:d72093d7faac4b94b752ce7ae445180ba286df851c4d3b7da514aa4ba5f8bafb
+    image: sphinxlightning/sphinx-relay:v2.2.5@sha256:65fd26fec704f3f9041b9ca60aeb5c47f4bcd5f644f295ee58afe526e6780bb0
     init: true
     restart: on-failure
     stop_grace_period: 1m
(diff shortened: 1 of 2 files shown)


apps/sphinx-relay/docker-compose.yml
```diff
@@ -3,7 +3,7 @@ version: "3.7"
 services:
   sphinx-relay:
     container_name: sphinx-relay
-    image: sphinxlightning/sphinx-relay:v2.2.3@sha256:d72093d7faac4b94b752ce7ae445180ba286df851c4d3b7da514aa4ba5f8bafb
+    image: sphinxlightning/sphinx-relay:v2.2.5@sha256:65fd26fec704f3f9041b9ca60aeb5c47f4bcd5f644f295ee58afe526e6780bb0
```
aphex3k, line 6: I ended up with the following digest:
```
sphinxlightning/sphinx-relay:v2.2.5@sha256:9c7b112ead4585a92b6d65ebaa6206cf64069813b382b361c6c928c0def8a682
```

lukechilds: Yeah this is the correct hash, just for future reference @Evanfeenstra you can do `docker pull sphinxlightning/sphinx-relay:v2.2.5` which will return the hash for the multiarch manifest.

The hash you've pinned to is just for the [amd64 image](https://hub.docker.com/layers/sphinxlightning/sphinx-relay/v2.2.5/images/sha256-65fd26fec704f3f9041b9ca60aeb5c47f4bcd5f644f295ee58afe526e6780bb0?context=explore) which will fail for all arm64 (Raspberry Pi) users.

Nw though I can quickly update this one before merge.

PR #1209, opened 2022-01-28
Update mempool app to v2.3.1

knorrium:
Version 2.3.1 fixes a bug that could cause data to be lost due to autocommit being disabled


wiz (approved):
ACK

lukechilds:
Thanks guys!

PR #1224, opened 2022-02-02
Upgrade ItchySats app to `0.4.0`

scratchscratchscratchy:
0.4.0 includes:
- Perpetual CFDs that are extended every hour.
- Some connection stability improvements.
- Some minor user interface improvements
- Internal model changes that require a hard reset of the CFD database. This is a breaking change.

Updated screenshots that reflect the interface changes:

1.
Screen Shot 2022-02-02 at 9 24 21 pm

2.
Screen Shot 2022-02-02 at 9 24 28 pm

3.
Screen Shot 2022-02-02 at 9 25 12 pm

### DIFF
--- a/apps/itchysats/docker-compose.yml
+++ b/apps/itchysats/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3.7"
 
 services:
   web:
-    image: ghcr.io/itchysats/itchysats/taker:0.3.6@sha256:14f2af91a1fc5a80e565c682680aa81a2c55e7a182268c1bef1de786993656b3
+    image: ghcr.io/itchysats/itchysats/taker:0.4.0@sha256:410332660845d7973c29a4a6b79b59b0748653dd73faade5b40e7c98061793ff
     restart: on-failure
     stop_grace_period: 1m
     ports:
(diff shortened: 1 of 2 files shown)


lukechilds:
When you say "This is a breaking change." do you mean the change is automatically handled by ItchySats or some manual steps are required by users?

scratchscratchscratchy:
> When you say "This is a breaking change." do you mean the change is automatically handled by ItchySats or some manual steps are required by users to upgrade?

It is automatically handled, no manual steps required. The only sad part is, that we lose the history of previous CFDs. The internal model change was so big that we were unable to migrate. In the future we are planning to always migrate the database.

The old version does not create offers anymore and we are closing all currently remaining positions to be closed using the oracle attestation. 
The new maker runs on a different port (we just updated that on the PR). Eventually we might change the old maker to run on an incompatible version so the users see a warning that they should upgrade (but they can't create new positions against the old version anyway).

lukechilds:
Got it, thanks for the explanation!

lukechilds:
Tested and working great, got a perpetual position open! 🚀

mayankchhabra:
Solid update! We [updated](https://github.com/getumbrel/umbrel-apps-gallery/commit/78fccd23840581f0f657cad3a2a2c12879170828) the gallery images with your latest screenshots 

The app icon and gallery images are directly fetched from [getumbrel/umbrel-apps-gallery](https://github.com/getumbrel/umbrel-apps-gallery) so if you ever want to tweak or update them, feel free directly open a PR on that repository. (Oh and if you need the Figma file just hit me up, cheers!)

PR #1239, opened 2022-02-08
Upgrade ItchySats app to `0.4.2`

scratchscratchscratchy:
0.4.2 includes:

- Bugfixes for wrongly showing CFD state when multiple network protocols are triggered together.
    - Fixes scenarios where and already closed CFD is shown as open in the user interface or a collaboratively close CFD is shown as force closed.
    - Deal with lock transactions being confirmed after a close (in case a user triggers closing the CFD while the lock transaction is still pending.
- Show the closing price in the UI once it is determined by a collaborative close or oracle attestation

### DIFF
--- a/apps/itchysats/docker-compose.yml
+++ b/apps/itchysats/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3.7"
 
 services:
   web:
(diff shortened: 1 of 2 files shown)


scratchscratchscratchy:
@lukechilds would be great to get these fixes in on top of the fixes we released in `0.4.1`. 
`0.4.2` fixes some scary UI display bugs 😬

lukechilds:
Just in time!

PR #1243, opened 2022-02-09
App Submission: Urbit Bitcoin Connector

yapishu:
# App Submission

### App name
Urbit Bitcoin Connector

...

### Version

0.1.0 

...

Use your Umbrel as a backend for your Urbit's Bitcoin wallet

...

### Summary of the app

Urbit has a native Bitcoin wallet that you can connect to a full node via an intermediary, called a provider. Providers run a piece of backend software called `%btc-provider` that communicates with a full node with `bitcoind` and `electrs`, as well as a custome Node Express proxy to translate RPC calls between TCP and HTTP. This app will install the Express proxy and configure it to point at your Umbrel's Bitcoin full node. Using this, you can configure a Bitcoin provider on a ship (even one running on your Umbrel) and configure it, or other ships to point their wallets at it.

This app also provides an information GUI web page, with instructions on connecting your provider ship and Bitcoin wallet. The web UI is accessible at port 9090. Note that if your Urbit is running on your Umbrel, you must configure your provider to point at the Umbrel's local/private IP address (like `http://192.168.0.10:55555`).

...

### Developer name

`~sitful-hatred`, `~mopfel-winrux`

Credit to `~timluc-miptev` & `~master-forwex`

...

### Developer website

[subject.network](https://subject.network)

...

### Source code link

[Github](https://github.com/yapishu/urbit-bitcoin-node)

...

### Support link

[Github Issues](https://github.com/yapishu/urbit-bitcoin-node)

`~matwet/networked-subject` on Urbit

...

### Requires
- [X] Bitcoin Core
- [X] Electrum server
- [ ] LND

### 256x256 SVG icon

...

### Gallery images

...

### I have tested my app on:
- [X] [Umbrel dev environment](https://github.com/getumbrel/umbrel-dev)
- [X] [Umbrel OS on a Raspberry Pi 4](https://github.com/getumbrel/umbrel-os)


apps/urbit-bitcoin-connector/docker-compose.yaml
```diff
@@ -0,0 +1,20 @@
...
+    environment:
+      ELECTRUM_IP: $ELECTRUM_IP
+      ELECTRUM_PORT: $ELECTRUM_PORT
+      BITCOIN_IP: $BITCOIN_IP
+      BITCOIN_RPC_PORT: $BITCOIN_RPC_PORT
+      BITCOIN_RPC_USER: $BITCOIN_RPC_USER
+      BITCOIN_RPC_PASS: $BITCOIN_RPC_PASS
+      BITCOIN_RPC_AUTH: $BITCOIN_RPC_AUTH
+    extra_hosts:
+        - "host.docker.internal:host-gateway"
```
lukechilds, line 20: What is this required for?

apps/urbit-bitcoin-connector/docker-compose.yml
```diff
@@ -0,0 +1,20 @@
+version: "3.7"
+
+services:
+  web:
+    image: matwet/urbit-btc-node:latest
```
lukechilds, line 5: This Docker image is only built for amd64. Umbrel apps need to use multiarch containers that support both amd64 and arm64. Most of Umbrel (>99%) is running on arm64.

It's pretty simple to support with `docker buildx`, we have an example here: https://github.com/getumbrel/umbrel/blob/master/apps/README.md#a-good-dockerfile

lukechilds (changes requested):

lukechilds:
Looking great at fist glance! We just need multiarch Docker images and ideally drop `host.docker.internal` which I don't think is needed unless I'm misunderstanding something.

apps/urbit-bitcoin-connector/docker-compose.yaml
lukechilds: Is this to simplify the connection between the "Urbit" and "Urbit Bitcoin Connector" apps? If so we can drop this and instead the apps can refer to each other via `$APP_URBIT_IP` (setup) and `$APP_URBIT_BITCOIN_CONNECTOR_IP` (not yet setup). This will use the IPs on the internal private Docker network.

yapishu: Yes, we were trying to figure out a static address we could put in the readme to copy/paste into the terminal witout success. An env var would be great!

yapishu:
Apologies @lukechilds , I thought the image was multiarch last time I checked, but the compose file is corrected and the image is pushed now.

apps/urbit-bitcoin-connector/docker-compose.yaml
lukechilds: Ah ok great!

Let me work on wiring that env var up, while I'm doing that do you want to tweak the container to accept an env var to write out to the instructions?

Once we're both done this should be good to merge and tag a new Umbrel release 🚀

yapishu: No problem!

lukechilds:
Just to clarify @yapishu you want the Urbit Bitcoin Connector's own IP passed to itself, right?

lukechilds:
@yapishu I think this should give you what you want: https://github.com/getumbrel/umbrel/pull/1243/commits/b742a4b2daf7fc135824796ab83a8308a3bd87fa. It'll be exposed inside the container as `$INTERNAL_IP` (easily changable if you prefer a different name).

That will be the IP of the Urbit Bitcoin Connector container on the internal private Docker network. The full Urbit app container is on the same network can ping this IP.

lukechilds:
Ahh one other thing I noticed is you have:

```
dojo> :btc-provider +bitcoin!btc-provider/command [%set-credentials api-url='http://local.ip.address.here:55555' network]
```

But the service is running on 55555 on the host and 50002 on the container. `$INTERNAL_IP` points to the container not the host, so we should change the container to also run on 55555.

yapishu:
@lukechilds Yes, we need the IP of the Bitcoin container (assuming the Urbit container can reach it) -- I'll try swapping out the var with $INTERNAL_IP.

I'll modify the script to use 55555 internally as well.

lukechilds:
Perfect!

yapishu:
Okay, it's pushed to Dockerhub, @lukechilds would you mind fixing the internal port and adding the $INTERNAL_IP var in the compose file?

lukechilds:
Already done!

lukechilds:
Let me give it one last test with the new image.

lukechilds:
@yapishu can you merge this? https://github.com/yapishu/urbit-bitcoin-node/pull/11

Should be working after this I think.

yapishu:
Okay, pushed

yapishu:
Nope, one more! exposed the wrong port in the dockerfile, just pushed it again

lukechilds:
Ahh that doesn't matter, the EXPOSE directive si kind of there as a self documenting thing, it doesn't actually do anything. (crazy I know)

This is tested and working great!

Screenshot 2022-02-09 at 22 00 24

Lets ship this bad boy! 🚀

mayankchhabra:
Congrats again on the launch, guys! [Here](https://github.com/getumbrel/umbrel-apps-gallery/tree/master/urbit-bitcoin-connector) are the gallery images and the app icon for Urbit Bitcoin Connector. If you ever want to change/update them, feel free to directly open a PR at [getumbrel/umbrel-apps-gallery](https://github.com/getumbrel/umbrel-apps-gallery). Cheers!

PR #1251, opened 2022-02-09
Umbrel v0.4.14

lukechilds:
This update brings four brand new apps to the Umbrel App Store — Urbit, Urbit Bitcoin Connector, Tailscale, and usocial, along with LND 0.14.2, updated apps, and Nunchuk wallet connection support.


mayankchhabra (approved):
hell yessss bois, what a release!

PR #1256, opened 2022-02-11
Upgrade usocial to v0.1.3.

ibz:
https://usocial.me/history#v0.1.3

Thank you!


lukechilds:
Tanks @ibz!

>Thinking whether there should be some script that just checks for new releases of all apps

Yeah we're aware the current app update process is not ideal haha. We have a cool solution in the works... 🤐

PR #1270, opened 2022-02-16
App Submission: Urbit Update

mopfel-winrux:
Updated Urbit app to version 0.2.0

-fixed a pier upload bug that was present at release.


apps/urbit/docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: "3.3"
 services:
   manager:
-    image: mopfelwinrux/urbit-umbrel:latest@sha256:72630ed72801d615a1ef5de0bbe0234855d7434d529542b09206a72fb4a8367c
+    image: mopfelwinrux/urbit-umbrel:latest
```
louneskmt, line 4: Ideally, you should pin a specific tag instead of `latest`.

louneskmt (commented):
Hey, thanks for the PR! FYI, there is no need to call it ˋApp Submission` and specify all the app details again. You can just name it `Update urbit app to X.X.X` and add some changelog.

mopfel-winrux:
Updated the docker-compsoe to include the sha and changed the PR comments to jsut be a change log.

lukechilds:
Thanks @mopfel-winrux!

mccolister1017:
still can't upload a keyfile , can't use it

mopfel-winrux:
@mccolister1017 can you submit a bug report to the urbit-umbrel repo? 
https://github.com/mopfel-winrux/urbit-umbrel

mccolister1017:
> @mccolister1017 can you submit a bug report to the urbit-umbrel repo? https://github.com/mopfel-winrux/urbit-umbrel

sure .

mopfel-winrux:
@lukechilds Any idea why this didn't get into the latest update?

lukechilds:
@mopfel-winrux Hmmn, are you sure you pushed your latest changes to Docker Hub?

I noticed hash you submitted in this PR didn't match the latest hash for the tag. You had submitted the checksum for the amd64 image, not the multiarch manifest, so this would have broken for all arm users. I updated it before merging: https://github.com/getumbrel/umbrel/pull/1270/commits/d6c5a95877155288436b27649dfc32550ac6d228

This is the correct hash for the multiarch manifest for the latest tag currently:

```
$ docker pull mopfelwinrux/urbit-umbrel:latest
latest: Pulling from mopfelwinrux/urbit-umbrel
Digest: sha256:72630ed72801d615a1ef5de0bbe0234855d7434d529542b09206a72fb4a8367c
Status: Image is up to date for mopfelwinrux/urbit-umbrel:latest
docker.io/mopfelwinrux/urbit-umbrel:latest
```


Double checking I now see that that is the same checksum we already had before this PR. Not quite sure what's gone wrong but it looks like maybe you pushed an identical image update on Docker Hub and then changed the image checksum from the multiarch manifest to the amd64 image in this PR.

Are you able to double check that your changes are definitely published to Docker Hub? Also I would suggest using versioned tags going forwards to make catching stuff like this a bit easier. So e.g `mopfelwinrux/urbit-umbrel:0.2.0` instead of `mopfelwinrux/urbit-umbrel:latest`.

To get the hash for the multiarch manifest you can run `docker pull <image>:<tag>`, near the end of the output you should see `Digest: sha256:<hash>`.

PR #1279, opened 2022-02-22
App Submission: Passky

zigazajc007:
# App Submission

### App name
[Passky](https://passky.org)

### Version
6.1.0

_(max 50 characters)_

Modern and open-source password manager

### Summary of the app
_(50 to 200 words)_

Passky is a simple, modern, lightweight, open source and secure password manager.

Passky is open-source. This means that anyone can download it, dig into its code, and customize it to their liking. Using Passky doesn’t require an upfront cost or monthly subscription. It’s completely free to download and can be hosted by anyone who has the space to do so.

Passky simplifies your digital life and solves your password problem. All you’ll ever need to do is remember a single, ultra-secure password that includes uppercase letters, lowercase letters, numbers, and special characters. Once you’ve got that written down in your wallet or purse (or buried in your memory), you’ll get the benefit of secure passwords across your online accounts without having to memorize dozens of codewords. This makes it one of the most secure ways to protect your online identity.

### Developer name
Rabbit Company LLC

### Developer website
https://passky.org
https://rabbit-company.com

### Source code link
_(Link to your app's source code repository.)_

https://github.com/Rabbit-Company/Passky-Server

### Support link
_(Link to your Telegram support channel, GitHub issues/discussions, support portal, or any other place where users could contact you for support.)_

Email: info@rabbit-company.com
Discord: https://discord.rabbit-company.com

Issues: https://github.com/Rabbit-Company/Passky-Server/issues

### Requires
- [ ] Bitcoin Core
- [ ] Electrum server
- [ ] LND

### 256x256 SVG icon
_(Submit an icon with no rounded corners as it will be dynamically rounded with CSS. GitHub doesn't allow uploading SVGs directly, so please upload your icon to an alternate service, like https://svgur.com, and paste the link below.)_

https://svgshare.com/i/eSC.svg

### Gallery images
_(Upload 3 to 5 high-quality gallery images (1440x900px) of your app in PNG format, or just upload 3 to 5 screenshots of your app and we'll help you design the gallery images.)_

Screenshot from 2022-02-22 15-54-07
Screenshot from 2022-02-22 15-55-40
Screenshot from 2022-02-22 15-56-09
Screenshot from 2022-02-22 15-56-14
Screenshot from 2022-02-22 15-56-22

## Question
Passky Server does not contain UI. That's why I have also included Passky Client in this application. Generally, it would be better for those 2 things to be separated, but if UI is required this should do the job.

### I have tested my app on:


lukechilds:
Thanks, will review this as soon as I'm able to!

lukechilds:
Have you tested this? I'm getting errors trying to run it, looks like the compose file is invalid.

># Question
>Passky Server does not contain UI. That's why I have also included Passky Client in this application. Generally, it would be better for those 2 things to be separated, but if UI is required this should do the job.

So the intended use case would be for users to run the Passky desktop client and connect to their Passky server, right? If so you could include a simple UI with the server app that just provides connection details and some simple instructions on how to connect a Passky client to it.

lukechilds (changes requested):
Have you tested this? I'm getting errors trying to run it, looks like the compose file is invalid.

zigazajc007:
> > # Question
> > Passky Server does not contain UI. That's why I have also included Passky Client in this application. Generally, it would be better for those 2 things to be separated, but if UI is required this should do the job.
> 
> So the intended use case would be for users to run the Passky desktop client and connect to their Passky server, right? If so you could include a simple UI with the server app that just provides connection details and some simple instructions on how to connect a Passky client to it.

Correct. As Passky has a really easy-to-use API, my goal is to bring more developers to create their own Clients. It's needed for every API call to return JSON format, so clients/developers can always fetch JSON format and return/check the output. If I would include a website (HTML, CSS) on the main domain, this would really confuse developers and also makes API kinda bloated as it's meant for Passky client to be written with under 1000 lines of code.

So less code = easier maintenance = potentially less bugs / vulnerabilities

It would be better to just wait for Umrel to drop the requirement for GUI.

Thanks so much for your time.

zigazajc007:
Simple UI has been included in Passky Server.

It will look like that: https://eu1.passky.org/

zigazajc007:
In the architecture scheme I can see that nginx (Proxy Manager is at the top), does that mean I don't need to include it in the application?

And for SSL / TLS certificates Umbrel will take care?

PR #1286, opened 2022-02-27
Add snowflake app UI

mayankchhabra:
Use a custom html file to wrap the GoTTY logs in an app-esque UI.


apps/snowflake/data/index.html
```diff
@@ -0,0 +1,130 @@
+<!doctype html>
+<html>
+  <head>
+    <title>Tow Snowflake Proxy</title>
```
nevets963, line 4: There is a typo here. This should be 'Tor'

apps/snowflake/data/index.html
```diff
@@ -0,0 +1,130 @@
...
+                </defs>
+            </svg>                                                  
+            <div class="app-details">
+                <span class="app-status success">&#9679; Running</span>
+                <h1 class="app-name">Tor Snowflake Proxy</h1>
+            </div>
+        </div>
+        <p class="text-muted">
+            Thank you for installing Tor Snowflake Proxy on your Umbrel to help defeat Tor censorship around the world.
```
nevets963, line 113: Isn't this helping to defeat censorship in general? Or is it only 'Tor censorship' ?

e.g. Thank you for installing the Tor Snowflake Proxy on your Umbrel to help defeat censorship around the world.

apps/snowflake/data/index.html
mayankchhabra: Oops, good catch!

apps/snowflake/data/index.html
mayankchhabra: Snowflake specifically is built to help people access the Tor network in countries where Tor is blocked. So while Tor helps defeat _censorship_, Snowflake helps defeat _Tor censorship_. Maybe @lukechilds can confirm?

PR #1297, opened 2022-03-01
[Snowflake] Add --title-format command

KayBeSee:
It looks like gotty adds its own title to the HTML template. Adding the --title-format explicitly sets the page title displayed in browser tabs.


lukechilds:
Great spot dude!

PR #1304, opened 2022-03-09
Urbit app update to v0.3.0

mopfel-winrux:
Fixed a bug with the last PR and added some information on the proper key format.


lukechilds:
Thanks @mopfel-winrux!

PR #1310, opened 2022-03-13
Update Bitfeed to v2.2.1

mononaut:
Changes:

- New transaction details view.
- Support for tiny screens (down to ~240x180).
- Backend stability and performance improvements.
- Fixes issue in v2.2.0 where fetching transaction inputs & fee data could break after a node restart


lukechilds:
Thanks @mononaut!

PR #1335, opened 2022-04-03
changed method of pulling hostname to be more universal

ClayShoaf:
I was getting errors when trying to install/run umbrel on an Arch system. It appears that `hostname` is an Ubuntu-specific command (or at least it doesn't exist in Arch). This should allow Umbrel to be compatible with more Linux systems, as long as they have all of the dependencies installed.

### DIFF
--- a/scripts/app
+++ b/scripts/app
@@ -106,7 +106,7 @@ compose() {
   local env_file="${UMBREL_ROOT}/.env"
   local app_base_compose_file="${UMBREL_ROOT}/apps/docker-compose.common.yml"
   local app_compose_file="${app_dir}/docker-compose.yml"
-  local app_domain="$(hostname -s 2>/dev/null || echo "umbrel").local"
+  local app_domain="$(cat /proc/sys/kernel/hostname -s 2>/dev/null || echo "umbrel").local"
   local app_hidden_servive_file="${UMBREL_ROOT}/tor/data/app-${app}/hostname"
   local app_entropy_identifier="app-${app}-seed"
 
--- a/scripts/app
+++ b/scripts/app
@@ -106,7 +106,7 @@ compose() {
   local env_file="${UMBREL_ROOT}/.env"
   local app_base_compose_file="${UMBREL_ROOT}/apps/docker-compose.common.yml"
   local app_compose_file="${app_dir}/docker-compose.yml"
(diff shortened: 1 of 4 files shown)


nevets963:
Thanks for the PR! I just took a look at the hierarchy to determine the hostname (as there is also `/etc/hostname`) and it looks like `/proc/sys/kernel/hostname` is most ideal.

Would you mind also patching up:
- https://github.com/getumbrel/umbrel/blob/master/scripts/app#L109
- https://github.com/getumbrel/umbrel/blob/master/scripts/update/01-run.sh#L165

ClayShoaf:
Alright, I think I did that right. Still figuring out the github interface. 

When I grepped, I should have left off the trailing parenthesis instead of doing `grep '$(hostname)'. That should be all of them now.

scripts/app
```diff
@@ -106,7 +106,7 @@ compose() {
   local env_file="${UMBREL_ROOT}/.env"
   local app_base_compose_file="${UMBREL_ROOT}/apps/docker-compose.common.yml"
   local app_compose_file="${app_dir}/docker-compose.yml"
-  local app_domain="$(hostname -s 2>/dev/null || echo "umbrel").local"
+  local app_domain="$(cat /proc/sys/kernel/hostname -s 2>/dev/null || echo "umbrel").local"
```
nevets963, line 109: It looks like you've left the `-s` from the hostname command

scripts/update/01-run.sh
```diff
@@ -162,7 +162,7 @@ if [[ "${current_umbrel_version}" = "0.4.0" ]] && [[ -f "${nextcloud_config_file
   echo
   echo "Fixing broken Umbrel v0.4.0 Nextcloud install..."
   nextcloud_hs=$(cat "${nextcloud_tor_file}")
-  nextcloud_local_url="$(hostname -s 2>/dev/null || echo "umbrel").local:8081"
+  nextcloud_local_url="$(cat /proc/sys/kernel/hostname -s 2>/dev/null || echo "umbrel").local:8081"
```
nevets963, line 165: As well as here re leaving the `-s`

scripts/app
ClayShoaf: Good catch. Thank you.

scripts/update/01-run.sh
ClayShoaf: They have been removed

PR #1359, opened 2022-04-22
Update lightning-terminal app to v0.6.7

guggero:
https://github.com/lightninglabs/lightning-terminal/releases/tag/v0.6.7-alpha

### DIFF
--- a/apps/lightning-terminal/docker-compose.yml
+++ b/apps/lightning-terminal/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3.7"
 
 services:
   web:
-    image: lightninglabs/lightning-terminal:v0.6.5-alpha@sha256:541fb9034a25dd4336becabdee6d40f76c3f4ae4557cbce30a46a3a8e3ef0e74
+    image: lightninglabs/lightning-terminal:v0.6.7-alpha@sha256:aba02fc2e41e9e2397d019e2d358e4bb0c518a80dd758bdfe888feeeb7aac28d
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
--- a/apps/registry.json
+++ b/apps/registry.json
@@ -430,7 +430,7 @@
         "id": "lightning-terminal",
         "category": "Lightning Node Management",
         "name": "Lightning Terminal",
-        "version": "v0.6.5-alpha",
+        "version": "v0.6.7-alpha",
         "tagline": "The easiest way to manage channel liquidity",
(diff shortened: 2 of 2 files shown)


lukechilds:
Thanks for this @guggero! Just to let you know we've seen it and it's all good but we're tweaking how apps works work a bit and it'll conflict with this PR so just wanna get that merged in and then we'll fix up the conflicts and merge this.

apps/lightning-terminal/docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: "3.7"
 
 services:
   web:
-    image: lightninglabs/lightning-terminal:v0.6.5-alpha@sha256:541fb9034a25dd4336becabdee6d40f76c3f4ae4557cbce30a46a3a8e3ef0e74
+    image: lightninglabs/lightning-terminal:v0.6.7-alpha@sha256:efeb992e9e1e55d4f48a65de11950d5b700674058b51043cc3e8ba0179b6ad34
```
nevets963, line 5: Can you change the digest to `sha256:aba02fc2e41e9e2397d019e2d358e4bb0c518a80dd758bdfe888feeeb7aac28d`?

The current digest only covers `amd64`, but the above covers `amd64` and `arm64`.

Otherwise this is good to be merged in.

nevets963 (changes requested):

apps/lightning-terminal/docker-compose.yml
guggero: Thanks @lukechilds for pushing the fixed checksum. I squashed the two commits.

Though I'm not sure I fully understand the difference between the two checksums (mainly because I'm not extremely familiar with how docker image manifests are structured).
I just want to avoid making the same mistake again next time. Do I understand it correctly that the _container image_ (`application/vnd.docker.container.image.v1+json`) digest should be used instead of the _distribution_ (`application/vnd.docker.distribution.manifest.v2+json`) digest?

That's the main difference I see between those two hashes:

```
"Ref": "docker.io/lightninglabs/lightning-terminal:v0.6.7-alpha@sha256:efeb992e9e1e55d4f48a65de11950d5b700674058b51043cc3e8ba0179b6ad34",
		"Descriptor": {
			"mediaType": "application/vnd.docker.distribution.manifest.v2+json",
			"digest": "sha256:efeb992e9e1e55d4f48a65de11950d5b700674058b51043cc3e8ba0179b6ad34",
			"size": 2011,
			"platform": {
				"architecture": "amd64",
				"os": "linux"
			}
		},
		"SchemaV2Manifest": {
			"mediaType": "application/vnd.docker.distribution.manifest.v2+json",
			"schemaVersion": 2,
			"config": {
				"mediaType": "application/vnd.docker.container.image.v1+json",
				"digest": "sha256:aeb697519a443d19973825ef5804781956f5858f6c39c4abfbd8de81b837afd1",
				"size": 3277
			},
			"layers": [ ... ]
		}
	}
]
```

lukechilds:
Thanks @guggero, sorry for the delay, this will be going out shortly!

apps/lightning-terminal/docker-compose.yml
louneskmt: I usually get the repo digest running this:
```
docker inspect $image --format {{.RepoDigests}} | sed -e 's/.*@//g' -e 's/]//g'
```

PR #1378, opened 2022-05-13
Elements: Add -assumvalid=tip

tiero:
Current version of Elements Core requires >8GB of RAM to verify confidential proofs during IBD.
 We skip historical block verification for now, as we don't have a way to set the memory limit
 
 Please @lukechilds review

### DIFF
--- a/apps/elements/docker-compose.yml
+++ b/apps/elements/docker-compose.yml
@@ -15,7 +15,7 @@ services:
       - -txindex=1
       # Current Elements Core requires >8GB of RAM to verify confidential proofs during IBD
       # We skip historical block verification for now, as we don't have a way to set the memory limit
-      - -assumevalid=tip
+      - -assumevalid=d4046e2c27e32e8582af39e402528e5df6ede1e1dd3c551764699e02e50bbf1e
       - -validatepegin=1
       - -fallbackfee=0.000001
       # Attach to Bitcoin network
--- a/apps/elements/docker-compose.yml
+++ b/apps/elements/docker-compose.yml
@@ -13,6 +13,9 @@ services:
     command:
       - -listen=1
       - -txindex=1
+      # Current Elements Core requires >8GB of RAM to verify confidential proofs during IBD
+      # We skip historical block verification for now, as we don't have a way to set the memory limit
+      - -assumevalid=tip
       - -validatepegin=1
       - -fallbackfee=0.000001
       # Attach to Bitcoin network


lukechilds:
Is `tip` a valid value for `assumevalid`? Whe I mentioned it in DMs I was implying `tip` as a placeholder for current tip block at the time of writing.

lukechilds:
>Current version of Elements Core requires >8GB of RAM to verify confidential proofs during IBD.

Are these expensive verification operations avoided by removing `validatepegin`? Or are they required just to verify normal transactions?

tiero:
> Or are they required just to verify normal transactions?

Yep they are required to check the "confidential transactions" are not inflating/burning assets. (the crypto magic is obscure to me, guess only Andrew and Sanket knows LOL)

Validation of pegin is actually easy, cause Bitcoin mainchain deposits and sebsequent issuance are always unconfidential.

lukechilds:
Ah ok got it.

In that case, I don't think assumevalid is gonna help. Because while we could hardcode it to the current chaintip, that only helps users who install directly after the Umbrel release. Any users who install a week from now will need to fully validate a full week of blocks and hit the memory issue.

Also I'd imagine even for users who install today, they skip the memory issue for IBD, but they'll run into it over time as cumulative memory builds up to validate incoming blocks right?

Maybe the best way to handle this for now is to ship as is currently, but put a caveat in the app store description that claims the Elements Core app is currently in beta and may result in high memory usage and instability?

What do you think?

tiero:
> Any users who install a week from now will need to fully validate a full week of blocks and hit the memory issue.

I guess it ramps up overtime. @dabloem did a test, hit that problem not that soon from genesis (from my memory it compounds after 1 year or so).

But it's true it means we need to push updates to the compose every month or so, so I understand If we do not want to commit to this.

> but they'll run into it over time as cumulative memory builds up to validate incoming blocks right

From my current experience after IBD, memory consumption to do block verification every minute is fine, I am stead at 0.2 GB RAM on my node

 > Elements Core app is currently in beta and may result in high memory usage and instability
 
 I think since it's pretty sure most of RPi will suffer for this fate and preventing sync, I would rather NOT ship it to be honest :(

lukechilds:
>(from my memory it compounds after 1 year of blocks or so).

Ah ok well that's pretty good.

Ok so lets set `assumevalid` to current block height then? Let say https://liquid.network/block/428848a3bcb1920a98da9f68144ff082e4ab1d5d20840aaeb655d02fe14e7ff1

And then that gives us plenty of time to come up with a more robust solution before memory usage becomes a problem.

tiero:
went one further :) [1831300](https://liquid.network/block/d4046e2c27e32e8582af39e402528e5df6ede1e1dd3c551764699e02e50bbf1e)

lukechilds:
Perfect!

PR #1381, opened 2022-05-13
Umbrel v0.4.18

lukechilds:
This update brings two brand new apps to the Umbrel App Store — Elements Core and IPFS Podcasting, along with updated apps and bugfixes.


mayankchhabra (approved):
let's goooo 💦🚀

PR #1427, opened 2022-06-09
Wrong command in CLI

kroese:
The command in the notice does not work, removing the "umbrel-" part fixes it.


louneskmt (approved):

nevets963 (approved):

lukechilds:
Thanks @kroese ❤️

PR #1440, opened 2022-06-12
Allow scripts/repo to use remote branches other than default

mononaut:
It's useful to be able to checkout specific branches of a development fork, to switch between different versions of an app for testing and troubleshooting.

Before the v0.5 release, you could do this via `scripts/update/update --repo <user>/<repo>#<branch>`, but the new "repo" script for umbrel-apps seems to be limited to the default branch of the given repository.

This PR adds an extra optional argument to that script to specify a branch.

e.g.
```bash
./scripts/repo set https://github.com/mononaut/umbrel-apps.git bitfeed-v2.3.3
./scripts/repo update
```

If the branch argument is omitted, the script uses the default branch of the repo as before.


nevets963:
Hey @mononaut 

In Umbrel 0.5.1, we added a `checkout` subcommand to make it fast and easy to pull specific branches from a remote repo. More details can be found here: https://github.com/getumbrel/umbrel-apps/issues/40#issuecomment-1184630460

I hope that helps!
Steve

Issue #1446, opened 2022-06-15
yq not available in Fedora, update script for 0.4.x tries to install using dpkg


Rudd-O:
Just wanted to note that the update script failed because it tried to install yq using a debian command in a fedora system.

Rudd-O:
Another problem:

```
/opt/umbrel/scripts/update/steps/migrate-to-repo.sh: line 23: dpkg-query: command not found
```

Rudd-O:
Actually this is comically poor quality:

https://github.com/getumbrel/umbrel/blob/master/scripts/update/steps/migrate-to-repo.sh#L23 just directly assumes we're running using Debian or Ubuntu.

This worked fine in Fedora when I was in version 0.4.x.

What gives?  Am I supposed to use Umbrel **only** with the operating system fork you guys produce?

nexus-uw:
i feel that  (quietly) dropping fedora support with only a minor update 0.4 -> 0.5 is rather rude

Ruppsn:
Same Problem here.
I know its suposed to be installed on Debian/Ubuntu but if you use docker you could make it distroless. 

sudo: apt-get: command not found 

Does someone have a solution? I was happy with umbrel till 0.5

Ruppsn:
Also tried the following:

`--no-install-docker --no-install-compose`

But same error

marzolfb:
This seemed to get me past the `sudo: apt-get: command not found` error on fedora

```
curl -L https://umbrel.sh | bash -s -- --no-install-docker --no-install-compose --no-install-deps
```

but then I still needed to install a bunch of dependencies manually that I didn't already have - fswatch, docker desktop, openssl.

PR #1447, opened 2022-06-15
Install/upgrade fixes.

Rudd-O:
Two circumstances prevent 0.5.0 from being successfully installed on Fedora.

These fixes should solve that problem.


Rudd-O:
Fixes https://github.com/getumbrel/umbrel/issues/1446 .

scripts/install
```diff
@@ -8,6 +8,7 @@ UPDATE_APT="true"
 INSTALL_APT_DEPS="true"
 INSTALL_AVAHI="true"
 INSTALL_YQ="true"
+if which yq >/dev/null 2>&1 ; then INSTALL_YQ="false" ; fi
```
aahmed7, line 11:
```suggestion
if ! command -v yq &> /dev/null ; then INSTALL_YQ="false" ; fi
```
`command` is better suited for such cases. avoid using `which`

aahmed7: Here is a good explanation: https://stackoverflow.com/a/677212

aahmed7 (changes requested):
replace `which` with the builtin `command`

Rudd-O:
Fixed.  Will rebase tomorrow.  Thanks.

Rudd-O:
Alright!  Rebased as promised, with a few quality improvements to boot.

scripts/update/01-run.sh
```diff
@@ -58,8 +58,9 @@ if [[ ! -z "${UMBREL_OS:-}" ]]; then
     fi
 
     # This makes sure systemd services are always updated (and new ones are enabled).
-    UMBREL_SYSTEMD_SERVICES="${UMBREL_ROOT}/.umbrel-${RELEASE}/scripts/umbrel-os/services/*.service"
-    for service_path in $UMBREL_SYSTEMD_SERVICES; do
+    UMBREL_SYSTEMD_SERVICES_ROOT="${UMBREL_ROOT}/.umbrel-${RELEASE}/scripts/umbrel-os/services"
+    UMBREL_SYSTEMD_SERVICES="*.service"
+    for service_path in "$UMBREL_SYSTEMD_SERVICES_ROOT"/$UMBREL_SYSTEMD_SERVICES ; do
```
aahmed7, line 63: it's better to do it like this:
`"$UMBREL_SYSTEMD_SERVICES_ROOT/$UMBREL_SYSTEMD_SERVICES"` 
the whole path is enclosed in quotes.

scripts/update/00-run.sh
```diff
@@ -16,8 +16,12 @@ echo
 
 # Make sure any previous backup doesn't exist
 if [[ -d "$UMBREL_ROOT"/.umbrel-backup ]]; then
-    echo "Cannot install update. A previous backup already exists at $UMBREL_ROOT/.umbrel-backup"
-    echo "This can only happen if the previous update installation wasn't successful"
+    backup_error="Cannot install update. A previous backup already exists at $UMBREL_ROOT/.umbrel-backup"
+    backup_error="$backup_error . This can only happen if the previous update installation wasn't successful."
+    echo "${backup_error}"
+    cat <<EOF > "$UMBREL_ROOT"/statuses/update-status.json
```
aahmed7, line 22: same here. better enclose the whole path in quotes. it's really just a cosmetic change and not something that would cause any problems

aahmed7 (dismissed):
added a few suggestions. otherwise looks good

scripts/update/01-run.sh
Rudd-O: Can't be done.  The point of doing it the way it currently is, is that you want the shell metacharacter expansion in the `for` loop.  If I quote as per your suggestion, that won't work anymore — thus defeating the point of the `for` loop.

scripts/update/00-run.sh
Rudd-O: This one can be done without breaking the functionality.  Stand by for fix...

aahmed7 (approved):
looks good

Rudd-O:
I can't merge.

nmfretz:
Thanks for this PR @Rudd-O. At this time, we are only officially supporting installation on Raspberry Pi and Ubuntu or Debian Systems. This fix would be a good thing to post on the [Community Forum](https://community.getumbrel.com/) to help others who wish to install on Fedora.

Rudd-O:
I don't get it.  This fix would not affect installation of Umbrel on Raspberry Pi or Ubuntu.  Why not merge it?

PR #1455, opened 2022-06-24
Add warning about wiping external drive

thebalaa:
Just lost all my data on an old drive. This would have helped me.

Cheers,

Mo


scripts/umbrel-os/status-server/static/index.html
```diff
@@ -89,6 +89,7 @@
       <div class="error-text no-block-device">
         <p><b class="text-dark">Error: No external drive found</b></p>
         <p>Please connect an external drive (at least 1TB) to a USB 3.0 port (blue color) on your Raspberry Pi and restart your Umbrel.</p>
+        <p><b class="text-dark">WARNING: The external drive will be wiped clean.</b></p>
```
mayankchhabra, line 92:
```suggestion
        <p>Note: If it's not an external drive that you have used with Umbrel previously, all of the data on it (if any) will be erased.</p>
```

mayankchhabra:
Oops, sorry about that @thebalaa! Thanks for the suggestion and the PR.

PR #1462, opened 2022-06-28
Add DEVICE_ID to exported environment vars

dngrhm:
Exports DEVICE_ID to the environment for `start` and `app` scripts.  The PR uses the same method already present for getting the device IP address.  This is helpful to ensure that the setup for client side applications can find the device on the network.  The hostname captured in the environement can be a bit finicky depending on the user's network.

### DIFF
--- a/scripts/app
+++ b/scripts/app
@@ -81,6 +81,7 @@ source_app() {
   export NETWORK_IP="${NETWORK_IP}"
 
   # Set other useful vars. used in exports
+  export DEVICE_IP="$(ip -o route get to 8.8.8.8 | sed -n 's/.*src \([0-9.]\+\).*/\1/p')"
   export DEVICE_HOSTNAME="$(cat /proc/sys/kernel/hostname 2>/dev/null || echo "umbrel")"
   export DEVICE_DOMAIN_NAME="${DEVICE_HOSTNAME}.local"
 
--- a/scripts/start
+++ b/scripts/start
@@ -83,6 +83,7 @@ if [[ -f "${UMBREL_ROOT}/tor/data/web/hostname" ]]; then
     hidden_service_url=$(cat "${UMBREL_ROOT}/tor/data/web/hostname")
     DEVICE_HOSTS="${DEVICE_HOSTS},http://${hidden_service_url}"
 fi
+export DEVICE_IP=$DEVICE_IP
 export DEVICE_HOSTS=$DEVICE_HOSTS
 export DEVICE_HOSTNAME="${DEVICE_HOSTNAME}.local"
 


nevets963:
Hi @dngrhm

Is there a particular usecase/app that you would use this with?

dngrhm:
I am getting ready to submit a couple apps that would add the Stacks blockchain, API, and explorer to Umbrel.

dngrhm:
I am no experiencing another reason for needing the DEVICE_ID.  I have multiple Umbrel servers on my network.  For my purposes, it is for testing Pi4 and x86_64 linux.  I could see someone running their local file/dev services on a separate Umbrel server than their Bitcoin server.  It is going to get messy if they are all called (and clients are looking for) umbrel.local.

dngrhm:
Hey @nevets963 

This feels stalled for a fairly small change.  Anything I can do to help move this along?

lukechilds:
Hey @dngrhm not sure I understand what problem this is solving, are you able to explain in more detail? So the idea is to pass the LAN IP of the Umbrel device to all apps, right? Could you explain what the problem is that you are facing and how this solves it?

I think this solution could be problematic for a few reasons. Finding the local IP is not always trivial and sometimes there can be multiple. e.g multiple network interfaces, overlay networks such as Tailscale, etc. How do we know which IP to return? Either could be the one you want in different scenarios. If we want to return all, it's also not always trivial to query all of these IPs. Some are owned at the OS level (ethernet), some are owned at the app level (Tailscale/WireGuard). Some network interfaces don't even have IPs (Tor). It's also not clear to me why an app needs to know its local IP. What problem does this solve? Also what if the IP changes while the app is running? The environment variable is now incorrect. We would need to restart all apps on any network interface changes to prevent env var data going stale. It doesn't feel like a good solution to me.

>I have multiple Umbrel servers on my network. For my purposes, it is for testing Pi4 and x86_64 linux. I could see someone running their local file/dev services on a separate Umbrel server than their Bitcoin server. It is going to get messy if they are all called (and clients are looking for) umbrel.local

Currently if someone is running two Umbrel's on the same network we recommend they set a custom hostname. So e.g you could have `umbrel-pi` and `umbrel-amd64` which are accessible at `http://umbrel-pi.local` and `http://umbrel-amd64.local` over mDNS or just `http://umbrel-pi` and `http://umbrel-amd64` over Tailscale with Tailscale MagicDNS enabled.

mayankchhabra:
Hey @dngrhm! Closing this PR for now due to inactivity. Please feel free to let us know if you think exporting `DEVICE_IP` is still required.

PR #1473, opened 2022-07-14
Umbrel v0.5.1

lukechilds:
This update brings several bugfixes, performance improvements, and the ability to see what’s changed in new app updates before installing them.


mayankchhabra (approved):
lesgoooo 🚀

PR #1499, opened 2022-09-08
scripts: remove "no issue detected" message in debug script

louneskmt:
The debug script will output this line no matter what happens. Users seeing this think there is no need to send the debug logs to me even though I asked for them, so support takes a few additional messages.

### DIFF
--- a/scripts/debug
+++ b/scripts/debug
@@ -186,5 +186,3 @@ if [[ ! -z "${UMBREL_OS:-}" ]]; then
         exit 0
     fi
 fi
-
-echo "The debug script did not automatically detect any issues with your Umbrel."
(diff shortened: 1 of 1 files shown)


mayankchhabra (approved):

mayankchhabra:
Shall we change it to:

```
====  END  =====
```

Because it logs "Result" right above it so it's awkward to end with no result.

nevets963:
I agree; if we print an `END` message, we have some confirmation that it's a complete debug log

mayankchhabra:
@louneskmt do you wanna update the script? We're ready to merge it then.

nmfretz:
Thanks @louneskmt! I added `==== END =====` as suggested by Mayank. Merged.

Issue #1476, opened 2022-07-16
i2p and bitcoin-core


awholegnuworld:
Currently the bitcoin-core implementation runs on IPv4, IPv6, and Onion. Will i2p be integrated? Screenshot of the Bitcoin RPC Explorer app:
![Screen Shot 2022-07-16 at 15 21 43](https://user-images.githubusercontent.com/56049227/179371168-9046a84b-da42-4b4b-9bd8-181eaf10a364.png)

nmfretz:
The Bitcoin Core app now has I2P integrated. You can manage connection settings in `Advanced Settings` within the app:

<img width="831" alt="image" src="https://github.com/getumbrel/umbrel/assets/85373263/74daf104-0c1c-4b43-9f24-a6f84e4991c0">

Issue #810, opened 2021-05-30
Allow i2p connections by default


dunxen:
Hey team! ☂️ 

It may be good to include an i2p daemon (such as `[i2pd](https://github.com/PurpleI2P/i2pd)`) and have the `bitcoin.conf` include i2p config by default (as it is replaced on every update). Since there are many umbrel users, this may help improve i2p connectivity across the network.

Relevant stack exchange answer: https://bitcoin.stackexchange.com/questions/103402/how-can-i-use-bitcoin-core-with-the-anonymous-network-protocol-i2p

AaronDewes:
See also https://github.com/getumbrel/umbrel/issues/694.

While I2P is a good idea, I think tor is enough for now, and the base Umbrel install should not include too many tools. But maybe this can be in the app store once we extend the app framework.

duncandean:
Thanks, I understand wanting to keep the base functionality with fewer tools. An app would be great once the framework is extended.

xanoni:
Would be trivial to add, no different from Tor. And I can confirm that it is picking up, around one third of my peers are I2P peers. A few months ago it was maybe 1 or 2.

count-null:
@AaronDewes Just wanting to revive this discussion in light of recent events. The UX of using tor to access node services has been degrading. I often have to wait multiple minutes or experience network timeouts due to Tor's increasing network congestion. 

Bitcoin core has enabled I2P support over 1-year ago. And there is a proposal in LND to [enable multiple proxies alongside Tor](https://github.com/lightningnetwork/lnd/issues/6845). There seems to be community support around the idea: [see this stacker news post](https://stacker.news/items/60428)

I'm hearing more node operators electing to abandon Tor-only due to these issues. In a time when online privacy is under attack, it feels like allowing multiple anonymity networks should be a priority.

dunxen:
@count-null, he doesn't work on this project anymore, but yeah now is a good time to consider i2p

nmfretz:
For those that find their way to this old issue, I2P connections are enabled by default now in the Bitcoin Node app. You can toggle this on and off in `Advanced Settings`

Issue #694, opened 2021-04-02
Add support for I2P network


kodxana:
Umbrella curently support only Tor network.
Would be nice if it would support I2P network so I could host and access my node over I2P.
More info on i2p: https://geti2p.net/en/

PR #1513, opened 2022-09-28
feat: add I2P

count-null:
First step towards fixing #694, #810, #1476

[Raspiblitz PR to add I2P](https://github.com/rootzoll/raspiblitz/pull/3330)

Adds i2pd docker image to core services. 
Next step is to modify bitcoin.conf to use the I2P proxy/router [as described in bitcoin-core](https://github.com/bitcoin/bitcoin/blob/master/doc/i2p.md)

By default, i2pd will operate as a router and passively transmit data from peers on the network. This could be a bandwidth issue for some users. 

[i2pd configuration](https://i2pd.readthedocs.io/en/latest/user-guide/configuration/) has options to limit bandwidth or disable transit. 

In the [i2pd Dockerfile](https://github.com/PurpleI2P/i2pd/blob/openssl/contrib/docker/Dockerfile#L14), configuration is hard coded as cli options. 

A PR is necessary in the i2pd repo to enable using a non-default configuration set. More in [this issue](https://github.com/PurpleI2P/i2pd/issues/1788)


awholegnuworld:
Amazing! i2p for the win!

awholegnuworld:
If this is added, would transmission also be able to connect over I2p? I can image this being useful for privacy in torrenting.

count-null:
> If this is added, would Transmission also be able to connect over I2p? I can image this being useful for privacy in torrenting.

It doesn't appear Transmission supports alternate data layers. May be possible with i2pd proxies. https://github.com/majestrate/XD is a torrent client which is i2p only. This would be a conversation for umbrel-apps repo tho. This PR only adds the I2P daemon to umbrel.

awholegnuworld:
Will the WebGui of I2p be open? It has lots of built in management features, and a torrent client.

count-null:
@awholegnuworld 
> Will the WebGui of I2p be open? It has lots of built in management features, and a torrent client.

This PR uses the lightweight i2pd implementation. The "official" implementation written in Java and includes a torrent client and management tools. The Java version is over 100MB and the lightweight i2pd (C++) binary is ~1MB.

Assuming i2p will be a core service, it should have a small footprint with functionality added in additional apps. 
[XD](https://github.com/majestrate/XD) has webUI and uses the SAM proxy (the same interface as bitcoin-core) so it should be a fine candidate for the app store once i2p is in umbrel. 

i2pd gets configured from file or command line args. There could be an umbrel setting for things like setting bandwidth limits, etc.

awholegnuworld:
Ah, neat! I've never used the lightweight client before! I knew of its existence, I just never tried it.

nevets963:
@count-null Many thanks for the PR!

Before Christmas, we updated the Bitcoin app to include the ability to use I2P. More details here: https://twitter.com/umbrel/status/1605906942555148288

With the above update - is it safe to now close this PR?

count-null:
Excellent work Umbrel team 👍

PR #1540, opened 2022-11-04
Append to instead of overwriting resolv.conf

river-jade:
Some networks do not have access to public DNS servers (such as Google or Cloudflare), and this breaks auto-update for Umbrel in a way that's very tricky to workaround.

The public DNS's are required for Umbrel users who are using Pinhole (in which case their router's DNS may be configured to point to the Umbrel instance for DNS services, which also breaks updates!).

To manage both of these contingencies, this commit appends the public DNS servers to the temporary resolv.conf, as well as setting the failover timeout to 1 second (instead of the default 5 seconds), so that non-pinhole users will simply use the original default, while pinhole users can just failover gracefully to the public ones.

### DIFF
--- a/scripts/update/00-run.sh
+++ b/scripts/update/00-run.sh
@@ -39,8 +39,9 @@ RESOLV_CONF_BACKUP_FILE="/tmp/resolv.conf"
 cat "${RESOLV_CONF_FILE}" > "${RESOLV_CONF_BACKUP_FILE}"
 
 # Use Cloudflare and Google public DNS servers for update
-echo "nameserver 1.1.1.1" > "${RESOLV_CONF_FILE}"
+echo "nameserver 1.1.1.1" >> "${RESOLV_CONF_FILE}"
 echo "nameserver 8.8.8.8" >> "${RESOLV_CONF_FILE}"
+echo "options timeout:1" >> "${RESOLV_CONF_FILE}"
 
 # Update status file
 cat <<EOF > "$UMBREL_ROOT"/statuses/update-status.json


scripts/update/00-run.sh
```diff
@@ -39,8 +39,9 @@ RESOLV_CONF_BACKUP_FILE="/tmp/resolv.conf"
 cat "${RESOLV_CONF_FILE}" > "${RESOLV_CONF_BACKUP_FILE}"
 
 # Use Cloudflare and Google public DNS servers for update
-echo "nameserver 1.1.1.1" > "${RESOLV_CONF_FILE}"
+echo "nameserver 1.1.1.1" >> "${RESOLV_CONF_FILE}"
 echo "nameserver 8.8.8.8" >> "${RESOLV_CONF_FILE}"
+echo "options timeout:1" >> "${RESOLV_CONF_FILE}"
```
nevets963, line 44: I like the general idea, but do you know the behaviour where they already have other DNS config? For example, say they have currently:

```
nameserver {umbrel_server_ip_address}
options timeout:2 rotate attempts:3
```

And then we append Public DNS servers as a backup, to become:
```
nameserver {umbrel_server_ip_address}
options timeout:2 rotate attempts:3
nameserver 1.1.1.1
nameserver 8.8.8.8
options timeout:1
```

Will the `options` simply override previous settings?

river-jade:
Great question! The documentation doesn't seem to specify it, but I had a
play around with it on Debian 10, and it looks like the last value is used,
so that's probably fine.

Interestingly, with the timeout set to 1, I get "ping: www.google.com:
Temporary failure in name resolution", but with it set to 2, it works ok.

Since this is somewhat of an edge case anyway (ie the original name servers
will work fine for most users, assuming that most users are not using
pi-hole for dns), I think that's probably ok.

Thoughts?

mayankchhabra:
Thank you for the PR, @river-jade! With the launch of [umbrelOS 1.0](https://github.com/getumbrel/umbrel/releases/tag/v1.0.0), we've introduced extensive changes to both the source code and the system's architecture. As part of this transition, we're reviewing and closing PRs and issues that might no longer be applicable to the new version. So I'll be closing this PR for now. Thanks again!

PR #1560, opened 2022-11-27
Increase the max Nginx client body size

ShonP40:
Fixes upload speed measuring on apps like OpenSpeedTest

More info can be found here: https://github.com/openspeedtest/Speed-Test/issues/23#issuecomment-1226781139

### DIFF
--- a/templates/nginx-sample.conf
+++ b/templates/nginx-sample.conf
@@ -30,6 +30,7 @@ http {
 			proxy_http_version 1.1;
 			proxy_set_header Upgrade $http_upgrade;
 			proxy_set_header Connection "upgrade";
+                        client_max_body_size 10000m;
 		}
 	}
 }


nevets963:
@ShonP40 Can you provide a bit more context on why this is needed? Nginx simply serves as a reverse proxy for Umbrel's Web UI and doesn't sit in front of any of your installed apps. We now have the App Proxy that does: https://github.com/getumbrel/umbrel/tree/master/deps/app-proxy which could potentially throttle bandwidth for an application like this.

ShonP40:
oh, didn't notice that it was a separate thing, my bad.

Well, is there a way to remove the throttle, or make it an option per app (or just a way to disable the proxy altogether per app)?

nevets963:
We don’t apply any throttling within the App Proxy. Are you seeing lower than expected performance with a speedtest app?

ShonP40:
It just doesn't even start testing upload (works fine without the proxy)

They have some more explanation for this here: https://github.com/openspeedtest/Speed-Test/issues/4#issuecomment-1229157193

nevets963:
I see. Do you have this app somewhere hosted in a Community App Store? Or how can I also quickly try this out?

ShonP40:
Yea, I have it here: https://github.com/ShonP40/umbrel-apps

openspeedtest:
@nevets963 Add OpenSpeedTest to  Community App Store? 
[Very simple docker image](https://hub.docker.com/r/openspeedtest/latest)
Docker image  : openspeedtest/latest:latest
port 3000 for http and 3001 for https.
That is it. 
[ScreenShots ](https://github.com/IceWhaleTech/CasaOS-AppStore/tree/main/Apps/OpenSpeedTest)
Warning! If you run it behind [a reverse proxy](https://github.com/openspeedtest/Speed-Test/issues/4#issuecomment-1229157193) you should increase post body content length to 35 Megabytes or more.

nevets963:
@ShonP40 @openspeedtest Ok, thanks guys! I'll run a test ASAP and report back

ShonP40:
Apparently deleting the forked repo closes the PR
Oops

openspeedtest:
@ShonP40  Installed umbrel for the First Time! It's AWESOME! Great UI/UX, OneClick App Install is SUPERB! 
openspeedtest
Tested your community app and got errors when upload test running.
@ShonP40  how to increase post body limit ?
@nevets963 Can't wait to see my app on the Official App Store :)

ShonP40:
> [@ShonP40](https://github.com/ShonP40) how to increase post body limit ?

that's why I made this PR lol

Can't figure out where Umbrel's app proxy is limiting it though.

openspeedtest:
Ohhh.. sorry. @ShonP40 I think it should work if we increase the limit to 35Mb or 100Mb.  If this is causing by Post Body limitation, otherwise we need to check buffering values.

nmfretz:
@openspeedtest I just tested this as well and see the same failed uploads as you show [here](https://github.com/getumbrel/umbrel/pull/1560#issuecomment-1336426137). Do you know what the repercussions of this are? Does it result in higher upload estimates than actual?

I'll need to dive a little deeper into this as I am not the creator of the App Proxy.

openspeedtest:
@nmfretz Use HTTP1.1 for OpenSpeedTest and raise the post body content length to 35 Mb or greater. The web server should wait for the upload request to finish before sending a 200 OK response. Otherwise, the application will perceive the upload request as completed, leading to inaccurate upload speed measurements. Essentially, the server is not behaving as desired. Therefore, modify the server's behavior or separate OpenSpeedTest onto a different IP or server.

ShonP40:
Well, I don't even know if this was fixed or not since I haven't been using Umbrel for over a year.

So just going to close this. (There have been many architectural changes either way so this PR wouldn't even work anymore)

PR #1561, opened 2022-11-27
Fix docker-compose output parsing

chokoboko:
This PR fixes the case where `docker-compose` `ps` command returns multiline output per container, resulting wrong names being passed down the line.

Tested on Ubuntu 20.04, x86, docker-compose 1.25.0

umbrl

### DIFF
--- a/scripts/memory-usage
+++ b/scripts/memory-usage
@@ -24,7 +24,7 @@ get_app_memory_use () {
 
   local app_memory=0
 
-  local app_containers=$("${UMBREL_ROOT}/scripts/app" compose "${app}" ps | awk '{print $1}' | grep -v 'Name\|-----')
+  local app_containers=$("${UMBREL_ROOT}/scripts/app" compose "${app}" ps | grep -vE '^\s' | awk '{print $1}' | grep -v 'Name\|-----')
   for container in $app_containers; do
     local container_memory=$(get_container_memory_use "${container}")
     app_memory=$(awk "BEGIN {print ${app_memory}+${container_memory}}")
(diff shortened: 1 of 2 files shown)


khoangkhac118 (approved):

khoangkhac118 (approved):

khoangkhac118 (approved):

khoangkhac118 (changes requested):
review-requested:[khoangkhac118]

nmfretz:
Hey @chokoboko. Thanks for this PR, and sorry for the delay in reviewing.

I see that in your example screenshot you are comparing your suggested change to this command that doesn't do any additional filtering and results in incorrect output:
`./scripts/app compose bitcoin ps | awk '{print $1}'`

But I think the right command to compare it to is actually this command that is currently being used in the script:
`./scripts/app compose bitcoin ps | awk '{print $1}' | grep -v 'Name\|-----'`

I get the same output that you show from your modified command:
```
./scripts/app compose bitcoin ps | awk '{print $1}' | grep -v 'Name\|-----'
bitcoin_app_proxy_1
bitcoin_bitcoind_1
bitcoin_i2pd_daemon_1
bitcoin_server_1
bitcoin_tor_1
```

Am I missing something?

mayankchhabra:
Thank you for the PR, @chokoboko! With the launch of [umbrelOS 1.0](https://github.com/getumbrel/umbrel/releases/tag/v1.0.0), we've introduced extensive changes to both the source code and the system's architecture. As part of this transition, we're reviewing and closing PRs and issues that might no longer be applicable to the new version. So I'll be closing this PR for now. Thanks again!

PR #1598, opened 2023-03-07
docs: add instructions to install Umbrel on Arch Linux

gutenye:


nmfretz:
Thanks for this PR @gutenye! At this time we are only officially supporting the installation methods currently shown in the README and our website (Raspberry Pi 4 and Ubuntu or Debian systems). This would be a great post to our [Community Forum](https://community.getumbrel.com/), so that other adventurous Arch Linux users can follow your instructions!

PR #1600, opened 2023-03-14
Give exitcode to systemd service

kroese:
Always when booting my system I get this error in my log:

`systemd[1]: umbrel-startup.service: Failed with result 'exit-code'.`

I have verified that the start script runs succesfully until the end, and the system is working fine. 

So my only guess is that systemd requires you to set the succesful exitcode explicitly, and this patch takes care of that.


nmfretz:
Hey @kroese, taking a look at this now. I can't seem to recreate this issue that you encounter. I see the following output after a successful boot:
` systemd[1]: Started Umbrel Startup Service.`

What are you running Umbrel on?

Issue #1558, opened 2022-11-23
Syntax error in root docker-compose.yml file preventing application specific logs and causing startup errors


cdmoss:
I came across this issue while trying to troubleshoot a bug preventing my lightning node app to not load. I attempted the method of checking application specific logs in [this forum post](https://community.getumbrel.com/t/how-to-generate-debugging-logs/5110) 
![image](https://user-images.githubusercontent.com/40777533/203646002-35e56128-7054-45b6-bf2e-7076a5c4623f.png)
but ran into an error that is described in #1370, which outputs this message in the log: `ERROR: The Compose file './docker-compose.yml' is invalid because:
networks.default.ipam.config.subnet is invalid: should use the CIDR format`.

 I peeked into the [docker-compose.yml](https://github.com/getumbrel/umbrel/blob/master/docker-compose.yml) file in the root umbrel directory, and found this syntactical error: 
![image](https://user-images.githubusercontent.com/40777533/203646555-d22229b5-f6e7-4424-8174-cf5c8c707690.png).
The value for `subnet` is missing curly braces, so the value of NETWORK_IP is not being substituted into the file. I added the curly braces, and now I can access the logs, and I presume that it might fix the issue I linked above. I can create a PR for this if the devs approve. :)

gacolitti:
>ERROR: The Compose file './docker-compose.yml' is invalid because: networks.default.ipam.config.subnet is invalid: should use the CIDR format

But your proposed solution does not resolve the error for me.

nevets963:
@cdmoss This is very strange and I've not seen this issue before. Can you confirm what version of `docker-compose` you have installed? Running `docker-compose --version` should return this.

@gacolitti Could you confirm if this file exists: `~/umbrel/.env`? You can run: `stat ~/umbrel/.env` and it should return with some data about the file, or return that the file does not exist.

gacolitti:
@nevets963 Unfortunately, I gave up on trying to fix the problem, wiped the SSD, and started from scratch.

JimSB:
This is also present in my umbrel/docker-compos.yml file, missing curly braces: 

![image](https://github.com/getumbrel/umbrel/assets/8608634/77b7c69b-0554-41c8-9590-16c4c0bd5cf1)

I will go ahead and make a PR to see what happens :)

nmfretz:
[...]

PR #1628, opened 2023-05-17
Adding curly braces to NETWORK_IP in config

JimSB:
Making PR to see if this is a thing, referenced here: https://github.com/getumbrel/umbrel/issues/1558

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -118,7 +118,7 @@ networks:
       ipam:
           driver: default
           config:
-              - subnet: "$NETWORK_IP/16"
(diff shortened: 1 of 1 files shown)


nmfretz:
Thanks @JimSB. It looks like the issue(s) encountered in https://github.com/getumbrel/umbrel/issues/1558 were likely unrelated to variable expansion. 

Docker Compose knows to substitute `NETWORK_IP` and not `NETWORK_IP/16` because the `/` character is not valid in environment variable names.

I will close related issues.

PR #1705, opened 2023-09-26
fix: blank app view in absence dependencies

ParthJadhav:


mayankchhabra (approved):
Awesome, thanks for the fix @ParthJadhav!

PR #1767, opened 2024-03-25
Add Ukrainian language

cor-bee:


cor-bee:
Updated all strings to umbrelOS 1.2 Beta 1

axbarchuk:
@cor-bee Is it possible to apply manually?

cor-bee:
> @cor-bee Is it possible to apply manually?

It must be, as any other Linux distro. 
However, I'm a newbie to test that myself, so I'm waiting for the umbrelOS team's response 😅

mayankchhabra:
Hi there @cor-bee. Thanks so much for your work. We've an automated GH Action to generate and update translations, and we [just included Ukranian](https://github.com/getumbrel/umbrel/blob/master/packages/ui/public/locales/uk.json) for our upcoming release. Although we can't accept submissions for the entire locale files for new languages, if you'd like to review the newly added translations and have any suggestions for changes, please feel free to open a PR against it. 🙏

cor-bee:
@mayankchhabra, happy to hear that!

I reviewed your Ukrainian translation and it looks good, nice job!

The only thing I disagree with is the translation for "app" as "додаток" (literally "add-on"). But since even big companies like Google still use that translation, I don't feel it's a big enough issue to rework a couple of dozen strings. 

Thank you again for adding the Ukrainian language to umbrelOS!

nmfretz:
Thanks for looking it over @cor-bee!

> The only thing I disagree with is the translation for "app" as "додаток" (literally "add-on"). But since even big companies like Google still use that translation, I don't feel it's a big enough issue to rework a couple of dozen strings.

It's really no added effort for us to change the translation for "app" if we've gotten it wrong. Is "програма" the more correct and commonly used translation for app in this case?

cor-bee:
@nmfretz, I would say that if we are talking about _computer_ apps, it's mostly called "програма".

However, there is another term - "застосунок", which is the literal translation of "app/application" and is used for _mobile_ apps like Telegram and Diia (Ukrainian digital ID app).

Since umbrelOS in a sense is not a full PC OS, I suggest that both variants "застосунок" (app) and "програма" (program) will do.

nmfretz:
Thanks for that explanation @cor-bee. Really appreciate it. Okay, we will go with "програма" based on what you used in this PR and because it is used in things like iOS/macOS so will be familiar to people.

PR #1780, opened 2024-04-02
Fix #1778

### LINKED ISSUE #1778: UmbrelOS install on Debian fails [solution found, needs review]
OS: Debian x64, just installed + installed `sudo` and `curl`

The UmbrelOS installation script fails while trying to install some pip libraries. This is caused by an update in pip, that disallows installing any packages through it, to protect the system (keyword: externally-managed-environment). If this system is deactivated by removing `/usr/lib/python3.11/EXTERNALLY-MANAGED` pip errors out with `AttributeError: cython_sources <...> error: subprocess-exited-with-error`.

highghlow:
This pull request fixed #1778

Commits:
- Revert scripts/install to v0.5.4: The installation script fetches the one from v0.5.4, so I had to revert the script to this version to edit it.
- Fix #1778:
  - Force bootstrapped: The script by default fetches itself from github which will overwrite the changes I made
  - Pin version: The current script passed --version to the old one. I can't do that here, so I pin the version in code
  - Modified install_docker_compose: https://github.com/getumbrel/umbrel/issues/1778#issuecomment-2029874353

The modified installation script has been tested on debian 12

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -1,7 +1,325 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+export DEBIAN_FRONTEND=noninteractive
+needrestart_conf_dir="/etc/needrestart/conf.d" 
+needrestart_conf_file="${needrestart_conf_dir}/temp-disable-for-umbrel-install.conf"
+sudo mkdir -p "${needrestart_conf_dir}"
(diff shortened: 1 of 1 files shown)


lukechilds:
Thanks so much for this @highghlow 🙏

Tested and working great, sorry for the trouble!

PR #1841, opened 2024-05-30
Fix HTTPS issue by using APP_PROTOCOL in axios baseURL

diazmateus:
This change fixes a compatibility issue with HTTPS when using the Cloudflare Tunnel to expose the Umbrel instance. The issue occurred because axios was configured with a fixed HTTP protocol. By modifying the configuration to use the protocol defined in the APP_PROTOCOL constant, all requests are now correctly made with HTTP or HTTPS as configured, preventing the browser from blocking requests when HTTPS is in use.

### DIFF
--- a/containers/app-proxy/utils/manager.js
+++ b/containers/app-proxy/utils/manager.js
@@ -4,7 +4,7 @@ const package = require('../package.json');
 const CONSTANTS = require('./const.js');
 
 const axiosInstance = axios.create({
-	baseURL: `http://${CONSTANTS.MANAGER_IP}:${CONSTANTS.MANAGER_PORT}`,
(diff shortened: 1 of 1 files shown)


aqaffaf (approved):

exileum (approved):

Exileum:
@nmfretz, can it be merged for the next OS release?

jackpurrin:
please merge umbrel devs

Anynomouss:
Not supporting https is major security weakness. Fix this ASAP! This should have been fixed years ago, or have been part of the initial release. In fact, if I knew there was not https support, I might not have bought an Umbrel home since it does not provide the level of security that I expect my devices to have.

Ari4ka (approved):

PR #1899, opened 2024-08-17
Language setting retrieval from local storage

JoseMoranUrena523:
In this version, I modify the `packages/ui/src/hooks/use-language.ts` to retrieve the language setting that was previously set in local storage.


JoseMoranUrena523:
Related to the README.md update, here is [a wiki page](https://github.com/JoseMoranUrena523/umbrel-wiki/blob/master/Install-umbrelOS-on-a-VPS.md) that I've created for VPS users.

PR #2161, opened 2026-05-30
fix: load docker-compose.override.yml in compose function

### LINKED ISSUE #2132: Bug: compose() in app-script does not load docker-compose.override.yml
## Problem

Docker Compose has a well-established convention: if a `docker-compose.override.yml` file exists alongside `docker-compose.yml`, it is automatically merged on top of the base config. This is Docker's standard mechanism for local customisations — adding volume mounts, overriding environment variables, exposing additional ports — without modifying the base file.

umbrelOS does not respect this convention. Users who place a `docker-compose.override.yml` in an app's data directory find it is silently ignored.

The practical consequence: the only way to add custom volume mounts to an app — for example, exposing a NAS share to Plex or the arr suite — is to directly edit `docker-compose.yml`. When an app update runs, `copy_app_files()` copies a fresh `docker-compose.yml` from the app store template, removing those edits with no warning.

---

## Steps to reproduce

1. Install any app from the Umbrel App Store (e.g. Plex)
2. Create a `docker-compose.override.yml` in the app's data directory
   at `~/umbrel/app-data/<app>/docker-compose.override.yml` with any
   valid volume mount or environment override
3. Restart the app via the Umbrel UI or
   `umbreld client apps.restart.mutate --appId <app>`
4. Run `docker inspect <container_name>` and observe that the
   override has not been applied

---

## Root cause

The `compose()` function in `packages/umbreld/source/modules/apps/legacy-compat/app-script` explicitly builds its own list of compose files using `--file` flags. Because compose is invoked with an explicit file list, Docker's automatic override detection is bypassed entirely. `docker-compose.override.yml` is never added to the file list.

```bash
# compose() — override file never included
compose_files+=( "--file" "${app_proxy_compose_file}" )
compose_files+=( "--file" "${tor_compose_file}" )
compose_files+=( "--file" "${common_compose_file}" )
compose_files+=( "--file" "${app_compose_file}" )
# docker-compose.override.yml is never added here

docker compose \
  --project-name "${app}" \
  "${compose_args[@]}"
```

---

## The override file already survives updates

The `pre-patch-update` command uses `copy_app_files()` with an explicit whitelist: `docker-compose.yml`, `*.template`, `exports.sh`, `torrc`, `hooks`. A `docker-compose.override.yml` placed in the app data directory is never touched by the update process. It already survives updates — it is simply never loaded.

---

## Proposed fix

Add 5 lines to the `compose()` function immediately after `app_compose_file` is added to `compose_files` (currently line 350), so the override is appended last and takes correct precedence:

```bash
# Add override file last if present, so it takes precedence
local -r override_compose_file="${app_data_dir}/docker-compose.override.yml"
if [[ -f "${override_compose_file}" ]]; then
  compose_files+=( "--file" "${override_compose_file}" )
fi
```

---

## Tested on a live install

The proposed fix was applied manually to `app-script` on a live umbrelOS install running Plex with a NAS share mounted at `/mnt/UNAS-PlexStorage`. After restarting Plex via `umbreld client apps.restart.mutate`, `docker inspect` confirmed the override volume was correctly applied. A simulated update (overwriting `docker-compose.yml` with the app store template) followed by a restart confirmed the override file survived and was applied correctly. No regressions observed on apps without an override file.

```
# docker inspect output after fix — mount correctly applied
"Source": "/mnt/UNAS-PlexStorage",
"Destination": "/mnt/UNAS-PlexStorage",
"Mode": "ro",
"RW": false,
"Propagation": "rprivate"
```

---

## Expected behaviour after fix

- A `docker-compose.override.yml` in an app's data directory is merged on top of the base compose config on every app start, stop, restart, and update
- The override file is never modified or removed by umbreld or the update process
- Custom volume mounts, environment overrides, and port additions persist across all app updates automatically
- If no override file is present, behaviour is identical to today

---

## Related

- #2043 — Feature request: native SMB/NFS share mounting
- umbrel-apps #3850 — [Jellyfin] Allow access to Videos path

ErnestHysa:
Fixes #2132 — compose() never loads docker-compose.override.yml.

This adds a check after the compose files are built to append the override file if it exists.

### DIFF
--- a/packages/umbreld/source/modules/apps/legacy-compat/app-script
+++ b/packages/umbreld/source/modules/apps/legacy-compat/app-script
@@ -349,6 +349,11 @@ compose() {
   compose_files+=( "--file" "${common_compose_file}" )
   compose_files+=( "--file" "${app_compose_file}" )
 
+  # Add override compose file if it exists
+  if [[ -f "${app_data_dir}/docker-compose.override.yml" ]]; then
+    compose_files+=(--file "${app_data_dir}/docker-compose.override.yml")
+  fi
+
   # Merge compose files and args. passed into 'compose'
   compose_args=("${compose_files[@]}" "${@}")
 


agent932 (approved):
Tested this fix on a live umbrelOS install — confirms the behaviour described in #2132.

Applied the patch manually to `/usr/local/lib/node_modules/umbreld/source/modules/apps/legacy-compat/app-script`, placed a `docker-compose.override.yml` in the Plex app data directory with a custom volume mount (`/mnt/UNAS-PlexStorage`), and restarted. `docker inspect` confirmed the override was applied correctly.

Also validated update survivability: overwrote `docker-compose.yml` with the app store template (simulating an app update), restarted, and the override was still picked up cleanly. Apps without an override file were unaffected.

LGTM — clean, minimal, correct.

DavidBurela:
I tried applying the patch locally as well.
Worked cleanly. The values in `docker-compose.override.yml` were correctly applied to the container.

jjmmbb:
When will this be released? A very simple and extremly important change.

agent932:
Hi @getumbrel, this PR addresses #2132 by enabling override compose files as documented. All tests pass and the change is minimal. Could you review when you have a chance? Thanks!

agent932:
@lukechilds is there anything else I need to do for this PR?

jjmmbb:
When will this fix be included in the UmbrelOS update?

PR #2187, opened 2026-07-09
fix: dont leave apps stuck in transient states when start/stop/restart fails

### LINKED ISSUE #2175: Container starts forever if compose syntax is wrong
Yeah
when you modify the compose files and you fuck up then it says starting and it will say starting for ever.
The only way to stop it is either uninstalling and reinstalling or to reboot your umbrel.

could you either implement something that allows restarting a what it thinks starting container or that it stops saying starting and just says failed or something.

Would be really nice because it sucks restarting it every single time because i made a typo
:(

ryanwaits:
Fixes #2175

## Problem

If an app fails to start, most commonly after hand-editing its `docker-compose.yml` and making a typo, the app gets stuck on "Starting..." forever. `App.start()` rejects after retries but `state` never leaves `'starting'`, so the UI spins indefinitely and the only ways out are uninstalling the app or rebooting, exactly as reported in #2175.

`stop()` and `restart()` have the same latent issue with `'stopping'` and `'restarting'`. The TODO above `AppState` in `app.ts` calls out this exact gap.

## Fix

**Revert the state to `'unknown'` on failure and re-throw**, mirroring what the boot path already does (`Apps.start()` sets `state = 'unknown'` when an app fails to start there). The UI already renders `'unknown'` as "Not running" with click-to-restart, so a user who breaks their compose file sees the app go offline, fixes the file, and clicks to bring it back. No reboot, no reinstall, no UI changes.

I kept this deliberately small. A dedicated `'failed'` state with its own UI treatment might be the fuller solution, but that touches the state schema and every UI switch over it. Happy to take that on as a follow up if you'd prefer it.

## Testing

Added an integration test: break an installed app's compose file with invalid YAML, verify `apps.start` rejects and state settles at `'unknown'` instead of sticking at `'starting'`, then restore the file and verify the app recovers to `'ready'`. The compose restore is in a `finally` so a future regression fails as one test instead of cascading through the sequential suite.

Verified in umbrel-dev (aarch64): the full apps integration suite passes 35/35 with this change, and the new test fails against master's `app.ts`. Also drove it manually in the browser: on master the app sticks on "Starting..." until a reboot; with this change it falls back to "Not running", and after fixing the compose file one click brings it back up.

## Caveat

Tested in the umbrel-dev VM (Docker/macOS), not on real umbrelOS hardware.

### DIFF
--- a/packages/umbreld/source/modules/apps/app.ts
+++ b/packages/umbreld/source/modules/apps/app.ts
@@ -213,18 +213,26 @@ export default class App {
 	async start() {
 		this.logger.log(`Starting app ${this.id}`)
 		this.state = 'starting'
-		// We re-run the patch here to fix an edge case where 0.5.x imported apps
-		// wont run because they haven't been patched.
-		await this.patchComposeFile()
-		await pRetry(() => appScript(this.#umbreld, 'start', this.id), {
-			onFailedAttempt: (error) => {
-				this.logger.error(
-					`Attempt ${error.attemptNumber} starting app ${this.id} failed. There are ${error.retriesLeft} retries left.`,
-					error,
-				)
-			},
-			retries: 2,
-		})
+		try {
+			// We re-run the patch here to fix an edge case where 0.5.x imported apps
+			// wont run because they haven't been patched.
+			await this.patchComposeFile()
+			await pRetry(() => appScript(this.#umbreld, 'start', this.id), {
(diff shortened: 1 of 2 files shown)


lukechilds:
Thanks for putting this together appreciate it! We've merged this into an internal dev branch for testing, it'll be included in the next release 🫡

PR #2188, opened 2026-07-09
fix: surface app start/stop/restart failures in the ui

ryanwaits:
## Problem

When an app fails to start, stop or restart, the tRPC mutation rejects but nothing in the UI listens for it: the mutations in `use-app-install.ts` have no `onError` handlers and there's no global mutation error toast. The failure is completely silent. The user just sees the app sitting in whatever state it landed in, with no idea anything went wrong or why (#2175 shows how confusing that gets).

## Fix

Show an error toast with the underlying message when start/stop/restart fails, following the existing error toast patterns (factory reset, files uploads). Three new strings in `en.json` only, assuming the translation workflow picks up the other locales. Happy to adjust if not.

The toast surfaces the failing command rather than a user-level explanation (e.g. "your compose file has invalid YAML"). Getting to that would need error shaping on the umbreld side, which felt out of scope for this change.

toast-failed-3

## Testing

Manually in umbrel-dev: broke an installed app's compose file, clicked the app, got "Failed to restart app" with the failing command as the description. The toast is selectable for copying, which the toaster's styling already anticipates.

## Caveat

Tested in the umbrel-dev VM (Docker/macOS), not on real umbrelOS hardware.

Pairs with #2187, which makes the app land in a recoverable state when this happens. The two are independent; either stands alone.


lukechilds:
Merged, same as https://github.com/getumbrel/umbrel/pull/2187#issuecomment-4933705365. Thanks!

